PR #12, opened 2016-09-21
Verify load of DAGMC plugin when starting Trelis via output to comman…

Pshriwise:
Upon opening Trelis, the string "Loading DAGMC export plugin..." is printed to the console for easy verification that the plugin has loaded.

### DIFF
--- a/export_dagmc_cmd/DAGMCExportCommand.cpp
+++ b/export_dagmc_cmd/DAGMCExportCommand.cpp
@@ -50,7 +50,7 @@ DAGMCExportCommand::DAGMCExportCommand() :
   if (console) {
     std::ostringstream load_message;
     load_message.str("");
-    load_message <<"Loading DAGMC export plugin..." << std::endl;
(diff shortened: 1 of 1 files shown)


Pshriwise:
It might seem odd that I used `print_error` rather than `print_message` but the former command was the only way I could get the string to appear in the Trelis console at load time.

Pshriwise:
Any comments on this?

kkiesling:
Does printing "Loading DAGMC export plugin..." indicate that it is currently loading (ie, a message should print that it was successful)? The "..." would make me think that the loading hung and it didn't actually work. Or should it instead be something like "Successfully loaded DAGMC export plugin." (ie, it has already successfully loaded and nothing else needs to be done)?

Pshriwise:
Mmm that's a fair point. I'm thinking something like "Loaded DAGMC export plugin."

gonuke:
Thanks @Pshriwise

PR #13, opened 2016-11-21
Add watertight opt

Pshriwise:
This PR adds and option to export a watertight mesh by adding `make_watertight` to the end of the export DAGMC Trelis command.

Optional expansion on current work:
 - check the mesh for watertight-ness and report success rate to user on file write

The current build process for this is to add `-DDAGMC_DIR=/path/to/dagmc/lib/` to mimic the one used for MOAB. It does not rely on a .cmake file in DAGMC currently as it does for MOAB includes and libraries.

### DIFF
--- a/export_dagmc_cmd/CMakeLists.txt
+++ b/export_dagmc_cmd/CMakeLists.txt
@@ -21,6 +21,8 @@ include_directories(${CUBIT_INCLUDE_DIRS})
 find_package(MOAB REQUIRED)
 include_directories(${MOAB_INCLUDE_DIRS})
 
+include_directories("/home/shriwise/dagmc_blds/dagmc/include/")
+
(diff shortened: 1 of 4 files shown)


Pshriwise:
Any thoughts on this?

ljacobson64:
I haven't really done any extensive testing, but the times I've tried it it seems to work

Pshriwise:
I think this is being tested empirically at ORNL right now if my UW box notifications are telling me the right story. Hopefully @gonuke and @makeclean will let me know if there are any problems.

makeclean:
So they turned out to be predominantly mac users, we dont currently have the mac version built, so it was best to have them all use the non mw version. Maybe I can try building the mac version later?

PR #14, opened 2016-12-17
Add prog bars

pshriwise:
Adds progress bars to the export command when curve faceting and surface faceting.

These would be nice to have for make_watertight, but the watertight is done doesn't lend itself as easily to adding them like it was here.


pshriwise:
Should I close this PR? Wait until we have better testing on this tool?

PR #31, opened 2017-04-21
MCNP Importer

ejwilson3:
Updates the plugin to call the library created by building the MCNP2CAD project in order to import MCNP format files directly into Trelis.


export_dagmc_cmd/CMakeLists.txt
```diff
@@ -1,9 +1,35 @@
 project(SDK_CommandPluginExample)
 
-# cmake_minimum_required(VERSION 3.1)
+ #cmake_minimum_required(VERSION 3.1)
```
pshriwise: Can this be removed?

export_dagmc_cmd/MyPlugin.cpp
```diff
@@ -10,7 +16,15 @@ MyPlugin::~MyPlugin()
...
   std::vector<std::string> keys;
+#ifdef BUILD_DAGMC_EXPORT
   keys.push_back("DAGMCExportCommand");
+#endif
+#ifdef BUILD_IGEOM_TESTS
+  keys.push_back("iGeom_test");
+#endif
+#ifdef BUILD_MCNP_IMPORT
+  keys.push_back("MCNPImp");
+#endif
```
pshriwise, line 28: I'm wondering if in the future it is going to be important to have each of these capabilities (or at least the importer/exporter) be entirely separate plugins registered with Trelis for the purpose of creating GUI elements that interact with their respective plugins. Perhaps we can cross that bridge when we come to it though.

ejwilson3: So far as I can tell, you'd want entirely different files (called in the examples "MyComp") when building the GUI elements, which create "components" which need to be loaded manually in Trelis.  The MyPlugin bit is focused on the command line capabilities.
I could be wrong, though.

gonuke:
I think this is now deprecated by my pushing a modified version to `develop` (based on all your commits, of course)

PR #32, opened 2017-07-05
Improve build instructions for mcnp2cad inclusion

gonuke:


pshriwise:
Nice. Exciting that the MCNP importer is almost ready!
Maybe this could go with the other caveats about building the MCNP importer below so they're all in one place?

gonuke:
@pshriwise : Do you meant that we should pull some of the README from the mcnp2cad repo?

pshriwise:
I meant move it below by the instructions about building without the MCNP importer, but after a re-read I think its best where it is now. 

**However**, I just noticed that the `-DMCNP2CAD_DIR=/path/to/libmcnp2cad` instruction. Can the `mcnp2cad` library/repo still exist anywhere or are we expecting it to exist in the DAGMC-Trelis location as a result of the clone instruction you've added?

I'd suspect that it can but I think you also mentioned some automated building of that library as part of the MCNP importer?

pshriwise:
Looks good to me. Thanks @gonuke!

PR #34, opened 2017-09-28
Update for curl and wget options

gonuke:
Also use the direct download URL


pshriwise:
Verified both of those options work for downloading the file with the direct URL. Thanks @gonuke!

PR #35, opened 2017-10-09
Updating install notes to include information about hdf5 and armadillo.

pshriwise:
Adding a couple of notes to the install steps to address dependencies of MOAB and mcnp2cad


pshriwise:
Anything that needs improving here?

PR #36, opened 2018-05-22
Add moab_setup.sh

SamuelStern:
Add a simple script to download, build, and install MOAB v5.0 and its dependencies. Future additions will refine this script and add scripts to sequentially build all dependencies of the DAGMC-Trelis plugin.

### DIFF
--- a/setup_scripts/moab_setup.sh
+++ b/setup_scripts/moab_setup.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+ 
+#MOAB Installation
+INSTALL_ROUTE=$HOME
+cd $INSTALL_ROUTE
+
+apt install -y autoconf libtool make mpich libblas-dev liblapack-dev libhdf5-dev cmake
+
+mkdir MOAB
+cd MOAB
+mkdir bld
+mkdir install
+git clone https://bitbucket.org/fathomteam/moab
+cd moab
+git checkout Version5.0
+autoreconf -fi
+cd ../bld
+../moab/configure --prefix=$INSTALL_ROUTE/MOAB/install --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial --enable-shared;
+make -j4
+make check
+make install
+printf 'export PATH=$PATH:%s/MOAB/install/bin' "$INSTALL_ROUTE">>/root/.bashrc
+export PATH=$PATH:$INSTALL_ROUTE/MOAB/install/bin
+printf 'export LD_LIBRARY_PATH=%s/MOAB/install/lib' "$INSTALL_ROUTE">>/root/.bashrc
+export LD_LIBRARY_PATH=$INSTALL_ROUTE/MOAB/install/lib
--- a/setup_scripts/moab_setup.sh
+++ b/setup_scripts/moab_setup.sh
@@ -6,17 +6,19 @@ cd $INSTALL_ROOT
 mkdir MOAB
 cd MOAB
 mkdir bld
-mkdir install
+MOAB_INSTALL_DIR=$INSTALL_ROOT/MOAB/install
+mkdir $MOAB_INSTALL_DIR
 git clone https://bitbucket.org/fathomteam/moab
 cd moab
 git checkout Version5.0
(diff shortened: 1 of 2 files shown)


setup_scripts/moab_setup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+ 
+#MOAB Installation
+INSTALL_ROUTE=$HOME
```
gonuke: this is more typically `ROOT` rather than `ROUTE`

setup_scripts/moab_setup.sh
```diff
@@ -0,0 +1,25 @@
...
+git clone https://bitbucket.org/fathomteam/moab
+cd moab
+git checkout Version5.0
+autoreconf -fi
+cd ../bld
+../moab/configure --prefix=$INSTALL_ROUTE/MOAB/install --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial --enable-shared;
+make -j4
+make check
+make install
+printf 'export PATH=$PATH:%s/MOAB/install/bin' "$INSTALL_ROUTE">>$HOME/.bashrc
```
gonuke: You can just use `$INSTALL_ROUTE` (or `$INSTALL_ROOT`) in the string like you did for `$PATH`

setup_scripts/moab_setup.sh
```diff
@@ -0,0 +1,25 @@
...
+mkdir MOAB
+cd MOAB
+mkdir bld
+mkdir install
+git clone https://bitbucket.org/fathomteam/moab
+cd moab
+git checkout Version5.0
+autoreconf -fi
+cd ../bld
+../moab/configure --prefix=$INSTALL_ROUTE/MOAB/install --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial --enable-shared;
```
gonuke: remove the `;`

gonuke (changes requested):
Thanks @SamuelStern - just a couple of changes

setup_scripts/moab_setup.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+ 
+#MOAB Installation
+INSTALL_ROUTE=$HOME
+cd $INSTALL_ROUTE
+
+apt install -y autoconf libtool make mpich libblas-dev liblapack-dev libhdf5-dev cmake
```
gonuke: Maybe move all the `apt` calls to a separate script because it needs to be run at `sudo` and the others don't

setup_scripts/moab_setup.sh
```diff
@@ -0,0 +1,22 @@
...
+git clone https://bitbucket.org/fathomteam/moab
+cd moab
+git checkout Version5.0
+autoreconf -fi
+cd ../bld
+../moab/configure --prefix=$INSTALL_ROOT/MOAB/install --with-hdf5=/usr/lib/x86_64-linux-gnu/hdf5/serial --enable-shared
+make -j4
+make check
+make install
+printf 'export PATH=$PATH:$INSTALL_ROOT/MOAB/install/bin'>>$HOME/.bashrc
```
gonuke: Might be better to print this out as guidance for the user to make this change, rather than making it automatically.

Also probably useful to prepend the new path since you want to look there first.

setup_scripts/moab_setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+ 
+#MOAB Installation
+INSTALL_ROOT=$HOME
+cd $INSTALL_ROOT
+mkdir MOAB
+cd MOAB
+mkdir bld
+mkdir install
```
gonuke: might want a `MOAB_INSTALL_DIR` variable or similar that can be reused below to ensure consistency, especially if/when you want to change it

gonuke (commented):
Looking good, just  couple of little things and then we can merge

gonuke (approved):

gonuke:
Looks good... thanks @SamuelStern

PR #37, opened 2018-05-25
MOAB Setup: Use eigen3

SamuelStern:
Adjust moab_setup.sh and apt_setup.sh to use cmake and eigen3 instead of Autotools, BLAS, and LAPACK.

### DIFF
--- a/setup_scripts/apt_setup.sh
+++ b/setup_scripts/apt_setup.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 #Install all simplest-level dependencies of DAGMC-Trelis that can be obtained via apt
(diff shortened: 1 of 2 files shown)


gonuke:
@SamuelStern - Did you mean to close this?

SamuelStern:
@gonuke No, I did not. Accidentally fumbled with some buttons on my fork, didn't realize it would immediately have this effect here.

gonuke:
Do you need to explicitly turn off BLASLAPACK? Or is it just good enough for CMake to fail to find them?

SamuelStern:
So far it appears that CMake failing to find them is sufficient.

gonuke:
For posterity, I think we want to explicitly avoid using BLASCHEK/LAPACK, so I think we should turn them off explicitly

gonuke:
looks good! thanks @SamuelStern

PR #38, opened 2018-06-09
Add dagmc setup

SamuelStern:
Adds dagmc_setup.sh to setup_scripts directory, for quick installation of DAGMC in a "fresh" development environment. Includes changes made in currently-open PR "MOAB Setup: Use eigen3"; it may be expedient to review those first.

### DIFF
--- a/setup_scripts/dagmc_setup.sh
+++ b/setup_scripts/dagmc_setup.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+#DAGMC Installation
+INSTALL_ROOT=$HOME
+cd $INSTALL_ROOT
+mkdir dagmc
+cd dagmc
+git clone https://github.com/svalinn/DAGMC
+cd DAGMC
+git checkout develop
+cd ..
+mkdir bld
+DAGMC_INSTALL_DIR=$INSTALL_ROOT/dagmc/install
+mkdir $DAGMC_INSTALL_DIR
+cd bld
+cmake ../DAGMC -DCMAKE_INSTALL_PREFIX=$DAGMC_INSTALL_DIR
+make 
+make install
+printf '\nConsider adding the following lines to .bashrc:\n'
+printf 'export PATH=$DAGMC_INSTALL_DIR/bin:$PATH\n'
+export PATH=$DAGMC_INSTALL_DIR/bin:$PATH
+printf 'export LD_LIBRARY_PATH=$DAGMC_INSTALL_DIR/lib:$LD_LIBRARY_PATH\n'
+export LD_LIBRARY_PATH=$DAGMC_INSTALL_DIR/lib:$LD_LIBRARY_PATH
--- a/setup_scripts/dagmc_setup.sh
+++ b/setup_scripts/dagmc_setup.sh
@@ -10,14 +10,17 @@ cd DAGMC
 git checkout develop
 cd ..
 mkdir bld
+
 DAGMC_INSTALL_DIR=$INSTALL_ROOT/dagmc/install
+MOAB_CONFIG_DIR=$INSTALL_ROOT/MOAB/install/lib/cmake/MOAB
 mkdir $DAGMC_INSTALL_DIR
 cd bld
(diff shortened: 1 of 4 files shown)


setup_scripts/dagmc_setup.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+#DAGMC Installation
+INSTALL_ROOT=$HOME
+cd $INSTALL_ROOT
+mkdir dagmc
+cd dagmc
+git clone https://github.com/svalinn/DAGMC
```
pshriwise: These few lines could be consolidated as `git clone -b develop https://github.com/svalinn/DAGMC` if you want

setup_scripts/dagmc_setup.sh
```diff
@@ -0,0 +1,26 @@
...
+cd DAGMC
+git checkout develop
+cd ..
+mkdir bld
+
+DAGMC_INSTALL_DIR=$INSTALL_ROOT/dagmc/install
+MOAB_CONFIG_DIR=$INSTALL_ROOT/MOAB/install/lib/cmake/MOAB
```
pshriwise: It might be good to have a common file to source these locations from. `INSTALL_ROOT` here might not be the same as it is in the `moab_setup.sh` script. If one changes but not the other, there could be problems.

setup_scripts/moab_setup.sh
```diff
@@ -9,16 +9,22 @@ mkdir bld
 MOAB_INSTALL_DIR=$INSTALL_ROOT/MOAB/install
 mkdir $MOAB_INSTALL_DIR
 git clone https://bitbucket.org/fathomteam/moab
```
pshriwise, line 12: Same comment as above, `git clone` provides a nice `-b` option for checking out a specific branch when cloning. Might be nice to use here.

setup_scripts/moab_setup.sh
```diff
@@ -9,16 +9,22 @@ mkdir bld
...
+
 cd moab
-git checkout Version5.0
```
pshriwise, line 13: The switch to `cmake` makes sense, but why are we not using Version5.0 anymore? I can't recall...

setup_scripts/dagmc_setup.sh
```diff
@@ -0,0 +1,26 @@
...
+mkdir $DAGMC_INSTALL_DIR
+cd bld
+cmake ../DAGMC -DCMAKE_INSTALL_PREFIX=$DAGMC_INSTALL_DIR -DMOAB_CMAKE_CONFIG=$MOAB_CONFIG_DIR
+make 
+make install
+
```
pshriwise: It would be good to run the DAGMC tests here as well. This is done by entering the `test` directory in the DAGMC installation directory. All executables in that location can be run using `run-parts .`

pshriwise (commented):
Nothing major. Small comments about variable consistency for easier maintenance down the line. Thanks @SamuelStern!

setup_scripts/moab_setup.sh
SamuelStern: The reason for this is that Version5.0 has no easy means of disabling BLAS or LAPACK in the installation. In order to install without either - the purpose of a lot of these changes - I'd either need to modify Version5.0, or just dodge it entirely. The latter was easier, but I admit I don't know much about the functional differences between that and the master branch.

setup_scripts/dagmc_setup.sh
SamuelStern: How should I go about that? For instance, should I simply add an "install_root_location" file to the setup scripts dir, and have the others call their install roots from that?

pshriwise: Something along those lines, yes. Maybe a file something like `install_locations.sh` and can be sourced (`source install_locations.sh`) in the various build script to keep the variables consistent.

setup_scripts/moab_setup.sh
pshriwise: Gotcha, that makes sense. Thanks!

PR #51, opened 2019-09-28
Fixed issues related to owenership of default tags

makeclean:
It seems some behaviour has changed in MOAB regarding the ownership of these 3 default tags. This change should be backwards compatible with other MOAB v5 releases but I havent tested it.


pshriwise:
Brilliant! Thanks for this. I've tested it with 5.1.0 and it works great. I'll merge it in on Monday.

PR #74, opened 2021-02-02
Linux action

bam241:
some test around auto plugin build for linux


script/CubitExport-release.cmake
```diff
@@ -0,0 +1,190 @@
+#----------------------------------------------------------------
+# Generated CMake target import file for configuration "Release".
+#----------------------------------------------------------------
```
gonuke: I think I asked this somewhere else, and I don't remember if it was answered.... is this file supposed to be in the SDK?  Wasn't it in the SDK in the past?

bam241: in this case nop, but in some yes. (small fix need to be done for macOS especially...)

bam241:
@Shimwell I can;'t get GitHub to pass my secret URL into the action, any thoughts about what I am doing wrong ?

bam241:
closing reopening from @svalinn

shimwell:
> @Shimwell I can;'t get GitHub to pass my secret URL into the action, any thoughts about what I am doing wrong ?

Perhaps a secret be added to the github repo then access it like this in the bash ```${{ secrets.MY_TOKEN }}```

PR #76, opened 2021-02-03
Linux action

bam241:
some test around GitHub action


.github/workflows/linux.yml
```diff
@@ -0,0 +1,157 @@
...
+          echo "CURRENT=$(pwd)" >> $GITHUB_ENV
+          echo "PLUGIN_ABS_PATH=" >> $GITHUB_ENV
+          echo "SCRIPTPATH=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          echo "PLUGIN_ABS_PATH=$GITHUB_WORKSPACE/.." >> $GITHUB_ENV
+
+      - name: Apt dependencies
+        shell: bash
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y g++ libtool libeigen3-dev libhdf5-dev patchelf git cmake
```
gonuke: Do we need `libtool`?

script/linux_share_build.sh
```diff
@@ -0,0 +1,121 @@
+#!/bin/bash
+
+
+
+function install_prerequise() {
+    TZ=America/Chicago
+    sudo ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+    sudo apt-get update -y
+    sudo apt-get install -y g++ libtool libeigen3-dev libhdf5-dev patchelf git cmake
```
gonuke: Same question about `libtool`

script/build_plug_linux.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
gonuke: Maybe some documentation on the arguments?

gonuke (commented):
I think this looks ok.  It's a lot of building of MOAB & DAGMC - can we eventually provide our own docker image?

.github/workflows/linux.yml
```diff
@@ -0,0 +1,153 @@
...
+    branches:
+      - develop
+      - master
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  main:
+    runs-on: ubuntu-18.04
```
pshriwise, line 20: If we can build this for multiple Ubuntu versions, we may as well do it. Unless we're confident that the plugin from 18.04 works for 20.04 of course.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,153 @@
...
+    name: 'Trelis Svalinn Plugin 17.1 Build'
```
pshriwise: Should this name have Linux, Ubuntu, and/or the OS version in it for clarity?

pshriwise (commented):
A couple of comments, but also a general question here. We have scripts for building the plugin for Linux and most of the lines here are copied from there. Can those scripts be refactored so that we can source the `linux_build_share.sh` file and call the functions from there for each step in this file? It would improve the maintenance outlook here I think.

.github/workflows/linux.yml
bam241: I think that is ok, as we ship most of the lib we depend on. (mainly hdf5...)

pshriwise: So the plugin built with Ubuntu 18.04 will work on 20.04?

bam241:
closing to be reopened in some other form

PR #78, opened 2021-02-08
Cubit 2020.2 GitHub actions

bam241:
copy of #72 from Svalinn

PR #81, opened 2021-02-15
Cubit 2020 compatibility

bam241:
this is an attempt to update the Plugin for compatibility with Trelis 17.1 and Cubit 2020.2


SvalinnPlugin.cpp
```diff
@@ -2,7 +2,7 @@
 #include <sstream>
 
 #include "CubitMessageHandler.hpp"
-#include "CubitInterface.hpp"
+#include "CubitCoreformInterface.hpp"
```
gonuke: What if this was something like:
```
#ifdef CUBIT_VERSION < XX.YY
#include "CubitInterface.hpp"
#define MESSAGE_HANDLER CubitInterface::get_cubit_message_handler()
#else
#include "CubitCoreformInterface.hpp"
#define MESSAGE_HANDLER CubitMessage::get_message_handler()
#endif
```
and below is something like:

``` 
CubitMessageHandler* console = MESSAGE_HANDLER;
```

bam241: I did something similar

but in a header file so it doesn't add to much copy paste...

bam241:
@pshriwise @gonuke this is ready to be reviewed,

gonuke:
This LGTM - do we have any actions or CI testing this?

bam241:
sadly not yet...

bam241:
I could merge it with #78  and then both would be tested

gonuke:
It probably makes sense to merge these - add build changes with testing changes

bam241:
closing there as it has been merged with #78

PR #83, opened 2021-02-17
Combining all the action

bam241:
This is not ready yet
But this aims to have a consistance set of action allowing to build all the version of the plugin of the multiple OS we are supporting


bam241:
closing to be reopened in some other form

PR #84, opened 2021-02-24
Cubit 2020.2 compatibility -- Linux action and scripts 

bam241:
This aims to:
1. modify the plugin implementation to build against Cubit 2020.2, see files :
     - `CMakeLists.txt`
     - `CubitVersionCompatibility.hpp`
     - `SvalinnPlugin.cpp`
     - `export_dagmc_cmd/DAGMCExportCommand.cpp`
     - `iGeom/iGeom.cpp`
     - ` iGeom/tests/iGeom_test.cpp`
     - `import_mcnp_cmd/MCNPImp.cp`
2. Update the build scripts to allow plugin build for Cubit 2020.2: `scripts/*.sh`
3. Setup GitHub Action to build both version of the plugin: `.github/workflows/linux.yml`


bam241:
> So the plugin built with Ubuntu 18.04 will work on 20.04?
@pshriwise to answer your question, I think so. I am not sure I have tried it, tho...

scripts/build_plugin_docker_ubuntu.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 SCRIPTPATH=`dirname $(dirname $(realpath $0))`
-docker run -v "$SCRIPTPATH:/Trelis-plugin" -v "$2:/Trelis-sdk" -it $1 bash -c "/Trelis-plugin/script/build_plugin.sh $3; bash"
+docker run -v "$SCRIPTPATH:/Trelis-plugin" -v "$2:/trelis-sdk" -it $1 bash -c "/Trelis-plugin/script/build_plugin.sh $3; bash"
```
gonuke: Would still love some documentation here of what $1, $2, $3 mean...

scripts/linux_share_build.sh
```diff
@@ -43,10 +46,9 @@ function build_dagmc(){
     cd ${PLUGIN_ABS_PATH}
     mkdir -pv DAGMC/bld
     cd DAGMC
-    git clone https://github.com/svalinn/DAGMC -b develop
+    git clone https://github.com/bam241/DAGMC -b build_exe
```
gonuke: Will this change once we merge DAGMC updates?

scripts/build_plugin_linux.sh
```diff
@@ -6,16 +6,21 @@ SCRIPTPATH=`dirname $(dirname $(realpath $0))`
 
 PLUGIN_DIR="plugin-build"
 PLUGIN_ABS_PATH=""
-
 #need to clear the LD_LIBRARY_PATH to avoid lib conflict
 unset LD_LIBRARY_PATH
+FOLDER_PKG="/trelis-sdk"
```
gonuke: This script only works in the docker build as launched in your other script, right?

gonuke (changes requested):
Just some Qs and documentation stuff

scripts/build_plugin_docker_ubuntu.sh
bam241: there is some int he `README.md` add some as comment in the file

gonuke (approved):

gonuke:
Once the required PR is in for Windows, should we do a micro-release of DAGMC and then use that tag for your build process?

bam241:
probably a good idea.
I think we probably need a micro release for PyNE, so we can use it for dagmc amalgamation...

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+    strategy:
+      matrix:
+        include:
+          - cubit: 17_1_0
+            os: ubuntu-18.04
+            URL: LINUX_CUBIT_17_1_0
+            URL_SDK: LINUX_CUBIT_SDK_17_1_0
+            
+          - cubit: 2020_2
+            os: ubuntu-20.04
```
pshriwise, line 30: Is there a particular reason that different Ubuntu versions are used for these?

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+          echo "CURRENT=$(pwd)" >> $GITHUB_ENV
+          echo "SCRIPTPATH=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          echo "PLUGIN_ABS_PATH=$GITHUB_WORKSPACE/.." >> $GITHUB_ENV
+          if [ "${{ matrix.cubit }}" = "17_1_0" ]; then
+            echo "CUBIT_PATH=/opt/Trelis-17.1" >> $GITHUB_ENV
+            echo "CMAKE_ADDITIONAL_FLAG=-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0" >> $GITHUB_ENV
+          else
+            echo "CUBIT_PATH=/opt/Coreform-Cubit-2020.2" >> $GITHUB_ENV
+          fi
+
```
pshriwise, line 86:
```suggestion
```

pshriwise:
Going to reiterate my comment from #76 here:

> We have scripts for building the plugin for Linux and most of the lines here are copied from there. Can those scripts be refactored so that we can source the linux_build_share.sh file and call the functions from there for each step in this file? It would be significantly easier to maintain one set of scripts than can be used in the GHA's as well.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Environment Variables
+        run: |
+          echo "CURRENT=$(pwd)" >> $GITHUB_ENV
+          echo "SCRIPTPATH=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          echo "PLUGIN_ABS_PATH=$GITHUB_WORKSPACE/.." >> $GITHUB_ENV
+          if [ "${{ matrix.cubit }}" = "17_1_0" ]; then
+            echo "CUBIT_PATH=/opt/Trelis-17.1" >> $GITHUB_ENV
+            echo "CMAKE_ADDITIONAL_FLAG=-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0" >> $GITHUB_ENV
```
pshriwise: I'm assuming this variable will be used to capture any additional CMake flags we pass to these commands in the future, so I'll suggest we make it plural

```suggestion
            echo "CMAKE_ADDITIONAL_FLAGS=-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0" >> $GITHUB_ENV
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+       
+      - name: Build MOAB
+        shell: bash
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          mkdir -pv moab/bld
+          cd moab
+          git clone https://bitbucket.org/fathomteam/moab -b Version5.1.0
+          cd bld
+          cmake ../moab $CMAKE_ADDITIONAL_FLAG \
```
pshriwise:
```suggestion
          cmake ../moab $CMAKE_ADDITIONAL_FLAGS \
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+
+      - name: build DAGMC
+        shell: bash
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          mkdir -pv DAGMC/bld
+          cd DAGMC
+          git clone https://github.com/svalinn/DAGMC -b develop
+          cd bld
+          cmake ../DAGMC $CMAKE_ADDITIONAL_FLAG \
```
pshriwise:
```suggestion
          cmake ../DAGMC $CMAKE_ADDITIONAL_FLAGS \
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+      - name: build plugin
+        shell: bash
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          cd Trelis-plugin
+          git submodule update --init
+          cd ../
+          mkdir -pv bld
+          cd bld
+          cmake ../Trelis-plugin $CMAKE_ADDITIONAL_FLAG \
```
pshriwise:
```suggestion
          cmake ../Trelis-plugin $CMAKE_ADDITIONAL_FLAGS \
```

scripts/build_plugin_linux.sh
```diff
@@ -1,21 +1,28 @@
...
 PLUGIN_DIR="plugin-build"
-PLUGIN_ABS_PATH=""
 
-#need to clear the LD_LIBRARY_PATH to avoid lib conflict
-unset LD_LIBRARY_PATH
+PLUGIN_ABS_PATH=""
+TRELIS_PATH=""
+TRELIS_PKG=""
+TRELIS_SDK_PKG=""   
+CMAKE_ADDITIONAL_FLAG=""
```
pshriwise:
```suggestion
CMAKE_ADDITIONAL_FLAGS="" 
```

scripts/linux_share_build.sh
```diff
@@ -29,10 +32,10 @@ function build_moab() {
             -DBUILD_SHARED_LIBS=ON \
             -DENABLE_BLASLAPACK=OFF \
             -DENABLE_FORTRAN=OFF \
-            -DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0 \
+            $CMAKE_ADDITIONAL_FLAG \
```
pshriwise:
```suggestion
            $CMAKE_ADDITIONAL_FLAGS \
```

scripts/linux_share_build.sh
```diff
@@ -56,36 +58,65 @@ function build_dagmc(){
                 -DBUILD_EXE=OFF \
                 -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
                 -DCMAKE_BUILD_TYPE=Release \
+                $CMAKE_ADDITIONAL_FLAG \
```
pshriwise:
```suggestion
                $CMAKE_ADDITIONAL_FLAGS \
```

scripts/linux_share_build.sh
```diff
@@ -56,36 +58,65 @@ function build_dagmc(){
...
-    dpkg -i Trelis-$1-Lin64.deb
+    if [ "$1" = "2020.2" ]; then
+        TRELIS_PATH="/opt/Coreform-Cubit-2020.2"
+        TRELIS_PKG="Coreform-Cubit-2020.2-Lin64.deb"
+        TRELIS_SDK_PKG="Coreform-Cubit-2020.2-Lin64-SDK.tar.gz"
+    elif [ "$1" = "17.1.0" ]; then
+        TRELIS_PATH="/opt/Trelis-17.1"
+        TRELIS_PKG="Trelis-17.1.0-Lin64.deb"
+        TRELIS_SDK_PKG="Trelis-SDK-17.1.0-Lin64.tar.gz"
+        CMAKE_ADDITIONAL_FLAG="-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0"
```
pshriwise:
```suggestion
        CMAKE_ADDITIONAL_FLAGS="-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0"
```

scripts/linux_share_build.sh
```diff
@@ -56,36 +58,65 @@ function build_dagmc(){
...
     cd Trelis-plugin
     git submodule update --init
     cd ../
     mkdir -pv bld
     cd bld
-    cmake ../Trelis-plugin -DCUBIT_ROOT=/opt/Trelis-${1::4} \
+    cmake ../Trelis-plugin -DCUBIT_ROOT=${TRELIS_PATH} \
                            -DDAGMC_DIR=${PLUGIN_ABS_PATH}/DAGMC \
                            -DCMAKE_BUILD_TYPE=Release \
+                            $CMAKE_ADDITIONAL_FLAG \
```
pshriwise:
```suggestion
                            $CMAKE_ADDITIONAL_FLAGS \
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+          rm -rf DAGMC/DAGMC DAGCM/bld
+
+      - name: Trelis setup
+        shell: bash
+        env:   
+          CUBIT_DEB_URL: ${{ secrets[matrix.URL] }}
+          CUBIT_SDK_URL: ${{ secrets[matrix.URL_SDK] }}
+        run: |
+          if [ "${{ matrix.cubit }}" = "2020_2" ]; then
+            sudo apt-get -y install libglew2.1 libglu1-mesa libopengl0 ocl-icd-libopencl1 libcxsparse3 libopenblas0 libegl1
```
pshriwise: Do we know why these libraries are needed for `Cubit-Learn 2020.2` but not `Trelis 17.1`?

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+          fi
+
+          cd ${PLUGIN_ABS_PATH}
+          curl -L https://uwmadison.box.com/shared/static/${CUBIT_DEB_URL}.deb --output trelis.deb
+          sudo dpkg -i trelis.deb
+
```
pshriwise:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+          sudo tar -zxvf trelis_sdk.tar.gz
+ 
+          if [ "${{ matrix.cubit }}" = "2020_2" ]; then
+            cd ${CUBIT_PATH}/bin
+            sudo cp -pv CubitExport.cmake CubitExport.cmake.orig
+            sudo sed -i "s/\"\/\.\.\/app_logger\"/\"\"/" CubitExport.cmake
+            sudo cp -pv CubitUtilConfig.cmake CubitUtilConfig.cmake.orig
+            sudo sed -i "s/\/\.\.\/app_logger\;//" CubitUtilConfig.cmake
+          fi
+
```
pshriwise:
```suggestion
```

.github/workflows/linux.yml
```diff
@@ -0,0 +1,186 @@
...
+          patchelf --set-rpath ${CUBIT_PATH}/bin/plugins/svalinn libpyne_dagmc.so
+          patchelf --set-rpath ${CUBIT_PATH}/bin/plugins/svalinn libuwuw.so
+
+          # Create the Svalinn plugin tarball
+          cd ..
+          ln -sv svalinn/libsvalinn_plugin.so .
+          cd ../..
+          tar --sort=name -czvf svalinn-plugin_linux_cubit_${{ matrix.cubit }}.tgz bin
+
+
```
pshriwise, line 121:
```suggestion
```

CubitVersionCompatibility.hpp
```diff
@@ -0,0 +1,14 @@
+#ifndef SVALINN_COMPATIBILITY_HPP
+#define SVALINN_COMPATIBILITY_HPP
+
+#include "CubitVersion.h"
+
+#if CUBIT_VERSION_MAJOR <= 17
```
pshriwise, line 6: Nice to have this compartmentalized. :+1:

scripts/build_plugin_docker_ubuntu.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
 
+
+# Usage: 
+# $1: docker image to use (tested with ubuntu:18.04 and ubuntu:20.04)
+# $2 path to local pkg folder containing both trelis/cubit install package and sdk
```
pshriwise: Any reason we can't automatically download the install packages and SDK from Box like we do in the GitHub actions?

scripts/build_plugin_linux.sh
```diff
@@ -1,21 +1,28 @@
 #!/bin/bash
 set -ex
 
+#need to clear the LD_LIBRARY_PATH to avoid lib conflict
```
pshriwise:
```suggestion
# need to clear the LD_LIBRARY_PATH to avoid lib conflict
```

scripts/linux_share_build.sh
```diff
@@ -56,36 +58,65 @@ function build_dagmc(){
...
+    elif [ "$1" = "17.1.0" ]; then
+        TRELIS_PATH="/opt/Trelis-17.1"
+        TRELIS_PKG="Trelis-17.1.0-Lin64.deb"
+        TRELIS_SDK_PKG="Trelis-SDK-17.1.0-Lin64.tar.gz"
+        CMAKE_ADDITIONAL_FLAG="-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0"
+    else
+        echo "unknown Trelis/Cubit version, use: \"17.1\" or \"2020.2\""
+    fi
+    if [ "$1" = "2020.2" ]; then
+        apt-get -y install libxt6 libglew2.1 libglu1-mesa libopengl0 ocl-icd-libopencl1 libcxsparse3 libopenblas0 libegl1 libboost-thread1.71.0 liblapack3
```
pshriwise: Some of these packages are present in the GHA ubuntu images, but not the docker ubuntu images I take it?

scripts/linux_share_build.sh
```diff
@@ -56,36 +58,65 @@ function build_dagmc(){
...
     cd /opt
-    tar -xzvf /Trelis-sdk/Trelis-SDK-$1-Lin64.tar.gz
+    tar -xzvf ${FOLDER_PKG}/${TRELIS_SDK_PKG}
+    if [ "$1" = "2020.2" ]; then
+        cd ${TRELIS_PATH}/bin
+        cp -pv CubitExport.cmake CubitExport.cmake.orig
+        sed -i "s/\"\/\.\.\/app_logger\"/\"\"/" CubitExport.cmake
+        cp -pv CubitUtilConfig.cmake CubitUtilConfig.cmake.orig
+        sed -i "s/\/\.\.\/app_logger\;//" CubitUtilConfig.cmake
+    fi
```
pshriwise: These regular expressions are one good example of why I think we should try to rely on the functions in these scripts in the GitHub actions and maintain only one version if possible.

Also, a comment as to what these lines are doing would be good.

pshriwise (changes requested):
Mostly formatting updates here. And I copied in that comment on the duplicate code in the GitHub actions and scripts. @bam241 if that is addressed in #85 and #86, just let me know and I won't let it hold up this work. I haven't looked at those PRs yet.

.github/workflows/linux.yml
bam241: Cubit 2020 is not compatible with package versions shipped with 18.04.

.github/workflows/linux.yml
bam241: I don't know why, just experience the dpkg complaining about those

scripts/build_plugin_docker_ubuntu.sh
bam241: yes we don't want to publicly expose the URL

scripts/linux_share_build.sh
bam241: maybe yes, 
I did not understood why one was working and not the other...

scripts/linux_share_build.sh
bam241: I agree and tried, but didn't found a way to satisfied me:
the problem is the GitHub Action job work on different terminal so local variable are not passed.

so one need to source the share_build at each steps
and when I update a variable value I need then to pass them to the system
I had the feeling that was getting too complicated.
(I though about removing the scripts, as the GitHub action were defined, but thought it is kinda nice to have them there.

Some thoughs
1.  maybe the complicated part could be extracted and shared between action and scripts.
2. maybe the docker script should not exist and the command line be mentioned in the README file

bam241: Here is a though,
I adapt the `line_shared_build.sh` script/method to look at the state of a variable that in `ON` if we are in GitHub action. When GIT_ACTION_ENV is `ON`, the value of variables are sent to the GitHub Action env variable handler.
it adds a bit of complexity in the build_script but allow sharing the methods 

What do you think @pshriwise ?

scripts/build_plugin_docker_ubuntu.sh
gonuke: Coreform may be moving towards a direct download  (per Shimwell)

.github/workflows/linux.yml
pshriwise: Just learning this today... if we use `apt` to install the Debian package the dependencies will be handled for us.

`sudo apt install -y ./Coreform-Cubit-2020.2-Ubuntu20.deb`

scripts/build_plugin_docker_ubuntu.sh
pshriwise: That would be great. Let's hope for that soon.

scripts/linux_share_build.sh
pshriwise: Hopefully the above comment about using apt will take care of this too.

scripts/linux_share_build.sh
pshriwise: Yeah there are a few different ways we can handle those issues I think. Can we separate the environment into a different script that always gets sourced and ensures the environment is setup for any of the build steps regardless of docker, local, and gha? The `GHA_ACTION_ENV` sounds idea fine to me.

gonuke: Is this critical path for this PR?
or can it be a follow up?
Or are we worried it won't happen that way?

pshriwise: I guess I think this is a good point to figure out how to rely on a common set of scripts that support all of these purposes. Otherwise, I think the code duplication in the GHA files is going to propagate faster than I'd like after looking ahead at #85 and #86.

If we handle this now, those PRs will be more focused on platform-specific issues than on new GHA files that look largely the same.

pshriwise: Maybe those will still look different enough that it isn't worth it though? @bam241 would know better than I. It just feels off seeing all of the same CMake/make commands reappearing in all of these new `.yml` files.

bam241:
running an export command with cubit 2020.2 and the new plugin I got:
```
Warning! ***HDF5 library version mismatched error***
The HDF5 header files used to compile this application do not match
the version used by the HDF5 library to which this application is linked.
Data corruption or segmentation faults may occur if the application continues.
This can happen when an application was compiled by one version of HDF5 but
linked with a different version of static or shared HDF5 library.
You should recompile the application or check your shared library related
settings such as 'LD_LIBRARY_PATH'.
You can, at your own risk, disable this warning by setting the environment
variable 'HDF5_DISABLE_VERSION_CHECK' to a value of '1'.
Setting it to 2 or higher will suppress the warning messages totally.
Headers are 1.10.4, library is 1.12.0
```

@gonuke @pshriwise shall we worry about it ?

bam241:
trying to export I also got a segfault:
here is the trace
```
#0  0x00007ffff5ed318b in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff5eb2859 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007fffe4f6152b in H5check_version (majnum=1, minnum=10, relnum=4) at H5.c:782
#3  0x00007fffa64e3b7a in mhdf_createFile () from /opt/Coreform-Cubit-2020.2/bin/plugins/svalinn/libMOAB.so.5
#4  0x00007fffa64b5519 in moab::WriteHDF5::serial_create_file(char const*, bool, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, moab::TagInfo* const*, int, int) () from /opt/Coreform-Cubit-2020.2/bin/plugins/svalinn/libMOAB.so.5
#5  0x00007fffa64a7c8a in moab::WriteHDF5::write_file_impl(char const*, bool, moab::FileOptions const&, unsigned long const*, int, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, moab::TagInfo* const*, int, int) () from /opt/Coreform-Cubit-2020.2/bin/plugins/svalinn/libMOAB.so.5
#6  0x00007fffa64a7207 in moab::WriteHDF5::write_file(char const*, bool, moab::FileOptions const&, unsigned long const*, int, std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, moab::TagInfo* const*, int, int) () from /opt/Coreform-Cubit-2020.2/bin/plugins/svalinn/libMOAB.so.5
#7  0x00007fffa6190a8f in moab::Core::write_file(char const*, char const*, char const*, moab::Range const&, moab::TagInfo* const*, int) () from /opt/Coreform-Cubit-2020.2/bin/plugins/svalinn/libMOAB.so.5
#8  0x00007fffa6190606 in moab::Core::write_file(char const*, char const*, char const*, unsigned long const*, int, moab::TagInfo* const*, int) () from /opt/Coreform-Cubit-2020.2/bin/plugins/svalinn/libMOAB.so.5
#9  0x00007ffff00a85b1 in DAGMCExportCommand::execute(CubitCommandData&) () from /opt/Coreform-Cubit-2020.2/bin/plugins/libsvalinn_plugin.so
#10 0x00007fffe92b3e79 in ?? () from /opt/Coreform-Cubit-2020.2/bin/libcubiti19.so
#11 0x00007fffe96b2a96 in ?? () from /opt/Coreform-Cubit-2020.2/bin/libcubiti19.so
#12 0x00007fffe92b6778 in ?? () from /opt/Coreform-Cubit-2020.2/bin/libcubiti19.so
#13 0x00007fffe9275c3a in ?? () from /opt/Coreform-Cubit-2020.2/bin/libcubiti19.so
#14 0x00007fffe92781f5 in ?? () from /opt/Coreform-Cubit-2020.2/bin/libcubiti19.so
#15 0x00007fffe9168b8e in CubitInterface::cmd(char const*) () from /opt/Coreform-Cubit-2020.2/bin/libcubiti19.so
#16 0x00007fffa5792fb1 in ?? () from /opt/Coreform-Cubit-2020.2/bin/_cubit2.so
#17 0x00007ffff126f32f in call_function (oparg=<optimized out>, pp_stack=0x7fffffffce20) at Python/ceval.c:4376
#18 PyEval_EvalFrameEx (f=f@entry=0x7ffff0f53210, throwflag=throwflag@entry=0) at Python/ceval.c:3013
#19 0x00007ffff12709c5 in fast_function (nk=<optimized out>, na=<optimized out>, n=1, pp_stack=0x7fffffffcf30, func=0x7fffa5dad9d0) at Python/ceval.c:4461
#20 call_function (oparg=<optimized out>, pp_stack=0x7fffffffcf30) at Python/ceval.c:4396
#21 PyEval_EvalFrameEx (f=f@entry=0x7ffff101c8c0, throwflag=throwflag@entry=0) at Python/ceval.c:3013
#22 0x00007ffff1271208 in PyEval_EvalCodeEx (co=co@entry=0x7ffff0f48230, globals=globals@entry=0x7ffff102d050, locals=locals@entry=0x7ffff102d050, args=args@entry=0x0, argcount=argcount@entry=0, kws=kws@entry=0x0, kwcount=0,
    defs=0x0, defcount=0, closure=0x0) at Python/ceval.c:3608
#23 0x00007ffff12713bd in PyEval_EvalCode (co=co@entry=0x7ffff0f48230, globals=globals@entry=0x7ffff102d050, locals=locals@entry=0x7ffff102d050) at Python/ceval.c:669
#24 0x00007ffff1296a8a in run_mod (arena=0x55555614aee0, flags=<optimized out>, locals=0x7ffff102d050, globals=0x7ffff102d050, filename=0x7ffff12c7175 "<string>", mod=<optimized out>) at Python/pythonrun.c:1385
#25 PyRun_StringFlags (str=0x7ffff1088f74 "cubit.cmd('export dagmc \"/home/mouginot/work/FRENSIE-tests/neutron_photon/bare_sphere/sphere.h5m\" faceting_tolerance 1.e-4')\n", start=start@entry=257,
    globals=globals@entry=0x7ffff102d050, locals=locals@entry=0x7ffff102d050, flags=<optimized out>) at Python/pythonrun.c:1348
#26 0x00007ffff126ad13 in exec_statement (locals=0x7ffff102d050, globals=0x7ffff102d050, prog=<optimized out>, f=0x7ffff0ff29b0) at Python/ceval.c:5126
#27 PyEval_EvalFrameEx (f=f@entry=0x7ffff0ff29b0, throwflag=throwflag@entry=0) at Python/ceval.c:2122
#28 0x00007ffff1271208 in PyEval_EvalCodeEx (co=<optimized out>, globals=<optimized out>, locals=locals@entry=0x0, args=<optimized out>, argcount=<optimized out>, kws=0x7ffff0ff33a8, kwcount=0, defs=0x0, defcount=0,
    closure=0x0) at Python/ceval.c:3608
#29 0x00007ffff126ddce in fast_function (nk=<optimized out>, na=<optimized out>, n=<optimized out>, pp_stack=0x7fffffffd320, func=0x7ffff101eed0) at Python/ceval.c:4471
#30 call_function (oparg=<optimized out>, pp_stack=0x7fffffffd320) at Python/ceval.c:4396
#31 PyEval_EvalFrameEx (f=f@entry=0x7ffff0ff3210, throwflag=throwflag@entry=0) at Python/ceval.c:3013
#32 0x00007ffff12709c5 in fast_function (nk=<optimized out>, na=<optimized out>, n=2, pp_stack=0x7fffffffd430, func=0x7ffff0f472d0) at Python/ceval.c:4461
#33 call_function (oparg=<optimized out>, pp_stack=0x7fffffffd430) at Python/ceval.c:4396
#34 PyEval_EvalFrameEx (f=f@entry=0x7ffff101c710, throwflag=throwflag@entry=0) at Python/ceval.c:3013
#35 0x00007ffff1271208 in PyEval_EvalCodeEx (co=co@entry=0x7ffff0ff47b0, globals=globals@entry=0x7ffff10f1170, locals=locals@entry=0x7ffff10f1170, args=args@entry=0x0, argcount=argcount@entry=0, kws=kws@entry=0x0, kwcount=0,
    defs=0x0, defcount=0, closure=0x0) at Python/ceval.c:3608
#36 0x00007ffff12713bd in PyEval_EvalCode (co=co@entry=0x7ffff0ff47b0, globals=globals@entry=0x7ffff10f1170, locals=locals@entry=0x7ffff10f1170) at Python/ceval.c:669
#37 0x00007ffff1296a8a in run_mod (arena=0x55555614e0c0, flags=0x7ffff10f1170, locals=0x7ffff10f1170, globals=0x7ffff10f1170, filename=0x7ffff12c7175 "<string>", mod=<optimized out>) at Python/pythonrun.c:1385
#38 PyRun_StringFlags (str=str@entry=0x555555fe6b28 "bhelper.instance().run_code('cubit.cmd(\\'export dagmc \"/home/mouginot/work/FRENSIE-tests/neutron_photon/bare_sphere/sphere.h5m\" faceting_tolerance 1.e-4\\')\\n')",
    start=start@entry=257, globals=0x7ffff10f1170, locals=0x7ffff10f1170, flags=flags@entry=0x0) at Python/pythonrun.c:1348
#39 0x00007ffff129868f in PyRun_SimpleStringFlags (
    command=0x555555fe6b28 "bhelper.instance().run_code('cubit.cmd(\\'export dagmc \"/home/mouginot/work/FRENSIE-tests/neutron_photon/bare_sphere/sphere.h5m\" faceting_tolerance 1.e-4\\')\\n')", flags=0x0)
    at Python/pythonrun.c:983
#40 0x00007ffff7e8594f in ?? () from /opt/Coreform-Cubit-2020.2/bin/libclarofw.so
#41 0x00007ffff7e8ca81 in ?? () from /opt/Coreform-Cubit-2020.2/bin/libclarofw.so
#42 0x00007ffff7e8b8d4 in ?? () from /opt/Coreform-Cubit-2020.2/bin/libclarofw.so
#43 0x00007ffff7e899e2 in Framework::run_app() () from /opt/Coreform-Cubit-2020.2/bin/libclarofw.so
#44 0x0000555555560c3b in ?? ()
#45 0x00007ffff5eb40b3 in __libc_start_main () from /lib/x86_64-linux-gnu/libc.so.6
#46 0x0000555555560f8e in ?? ()
```

bam241:
wondering if the 2 are related

pshriwise:
> shall we worry about it?

Yes. We should worry about it. First step is to figure if this is a clash between the hdf5 libraries shipped with Cubit and the one used to build DAGMC I think.

> running an export command with cubit 2020.2 and the new plugin I got:

What Ubuntu version?

bam241:
> What Ubuntu version?

20.04

bam241:
> Yes. We should worry about it. First step is to figure if this is a clash between the hdf5 libraries shipped with Cubit and the one used to build DAGMC I think.

I'll try to uninstall my hdf5 package and see what append.

bam241:
removing all hdf5 apt packages from my computer, I still get the warning and the segfault.

pshriwise:
we include the hdf5 libraries with the plugin, right? So it's still going to find those.

pshriwise:
Good to see all the work going into relying on our bash scripts for GHA. I'm going unsubscribe from the updates here until the work stabilizes a bit. Feel free to ping me when this is ready for another look @bam241!

bam241:
@pshriwise @gonuke this seems to be working properly (only tested cubit 2020.2 build on githubaction on my local desktop)
what I did so far:
- added GHA variable and when "ON" upload the script variable to GITHUB_ACTION ENV system
- update the GitHub action linux.yml to leverage the shared_script
- introduced a `$SUDO`variable empty by default and set to `sudo` to allow run some command as `sudo` in the GitHub Action
- add a function to install hdf5 1.12.0 from source
- had to implement an hot fix on MOAB CMakefile.txt to `find(HDF5)` instead of `find(HDF5_MOAB)` similar as what I needed for windows... (I think that can live as is for now... and we try to push that change into MOAB in the future)

I didn't find a way to make this work without the tweak on MOAb CMakefile for HDF5, I didn't investigated more, but it seems to me that this `FindHDF5_MOAB.cmake` shipped with MOAB source code as problems in it....

bam241:
I just tested the Trelis17.1 plugin it seems to be working as well

Both have been tested on: https://github.com/bam241/FRENSIE-tests/blob/neutron_photon/neutron_photon/bare_cube/trelis_script_pb

.github/workflows/linux.yml
```diff
@@ -0,0 +1,119 @@
...
+          cd $FOLDER_PKG
+          echo ${TRELIS_PKG}
+          curl -L https://uwmadison.box.com/shared/static/${CUBIT_DEB_URL}.deb --output ${TRELIS_PKG}
+          curl -L https://uwmadison.box.com/shared/static/${CUBIT_SDK_URL}.gz --output ${TRELIS_SDK_PKG}
+
+
+      - name: Apt dependencies
+        shell: bash
+        run: |
+          source $GITHUB_WORKSPACE/scripts/linux_share_build.sh
```
pshriwise: These repeated sources of the `linux_share_build.sh` file are a bit of an annoyance.

It looks like there may be a way around this 

https://github.community/t/how-to-share-shell-profile-between-steps-or-how-to-use-nvm-rvm-in-steps/16398

CMakeLists.txt
```diff
@@ -80,6 +71,8 @@ if (BUILD_IGEOM)
   include_directories(iGeom)
   add_library(iGeom SHARED ${IGEOM_SRC_FILES})
   target_link_libraries(iGeom ${CUBIT_LIBS})
+   target_include_directories(iGeom PUBLIC
+              $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
```
pshriwise:
```suggestion
   target_include_directories(iGeom PUBLIC
                              $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>)
```

scripts/linux_share_build.sh
```diff
@@ -3,33 +3,82 @@ PROC=$((`grep -c processor /proc/cpuinfo`))
 
 function install_prerequisites() {
     TZ=America/Chicago
-    ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
-    apt-get update -y
-    apt-get install -y g++ libeigen3-dev libhdf5-dev patchelf git cmake
+    $SUDO ln -snf /usr/share/zoneinfo/$TZ /etc/localtime
```
pshriwise, line 6: We should be able to install the `sudo` command in the docker images, right?

scripts/linux_share_build.sh
```diff
@@ -56,36 +104,51 @@ function build_dagmc(){
...
                 -DCMAKE_BUILD_TYPE=Release \
+                $CMAKE_ADDITIONAL_FLAGS \
                 -DCMAKE_INSTALL_PREFIX=${PLUGIN_ABS_PATH}/DAGMC
+                
+    
     make -j$PROC
     make install
     cd ../..
-    rm -rf DAGMC/DAGMC DAGMC/bld
+    rm -rf DAGMC/DAGMC DAGCM/bld
```
pshriwise:
```suggestion
    rm -rf DAGMC/DAGMC DAGMC/bld
```

scripts/linux_share_build.sh
```diff
@@ -3,33 +3,82 @@ PROC=$((`grep -c processor /proc/cpuinfo`))
...
+        TRELIS_PATH="/opt/Trelis-17.1"
+        TRELIS_PKG="Trelis-17.1.0-Lin64.deb"
+        TRELIS_SDK_PKG="Trelis-SDK-17.1.0-Lin64.tar.gz"
+        CMAKE_ADDITIONAL_FLAGS="-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0"
+    else
+        echo "unknown Trelis/Cubit version, use: \"17.1\" or \"2020.2\""
+    fi
+
+    # upload the variable in GithubAction
+    if [ "$GHA" = "ON" ]; then
```
pshriwise: Similar to the comment above, I wonder if modifying the `.bash_profile` might be an option for setting a lot of these environment variables in GHA.

scripts/linux_share_build.sh
```diff
@@ -3,33 +3,82 @@ PROC=$((`grep -c processor /proc/cpuinfo`))
...
     cd moab
     git clone https://bitbucket.org/fathomteam/moab -b Version5.1.0
+    cd moab
+    # patching MOAB CMakeLists.txt to use default find(HDF5)
+    sed -i "s/HDF5_MOAB/HDF5/" CMakeLists.txt
+    cd ..
+    #end of patch
     cd bld
     cmake ../moab -DENABLE_HDF5=ON \
-            -DCMAKE_PREFIX_PATH=/usr/lib/x86_64-linux-gnu/hdf5/serial \
```
pshriwise, line 28: Were you having trouble finding HDF5 in docker or GHA? I'm curious if I might be able to try a few things out locally.

pshriwise (changes requested):
Just a few comments/suggestions here. It's looking really good. My hope is that all this work up front will make integration of other builds really smooth :crossed_fingers:

scripts/linux_share_build.sh
shimwell: Yep I use this elsewhere

```RUN apt-get --yes install sudo```

File: scripts/linux_share_build.sh (hunk 29 above)
Comment (bam241) [reply scripts/linux_share_build.sh:28]: both as soon as I install my own hdf5, I can't use the `find(HDF5_MOAB)` from the MOAB source

File: scripts/linux_share_build.sh (hunk 26 above)
Comment (bam241) [reply scripts/linux_share_build.sh:6]: yes we could I was just scared of the potential effect of using sudo there, as those script are also there for direct use...

File: .github/workflows/linux.yml (hunk 24 above)
Comment (bam241) [reply .github/workflows/linux.yml]: I'll take a look!

thx for the intel!!

File: scripts/linux_share_build.sh (hunk 26 above)
Comment (pshriwise) [reply scripts/linux_share_build.sh:6]: Ah yeah that's a good point. Nevermind.

File: scripts/linux_share_build.sh (hunk 29 above)
Comment (pshriwise) [reply scripts/linux_share_build.sh:28]: Strange. I wasn't having any problems on my Ubuntu install with a custom hdf5 build. I'll try it out in docker and see what I get.

Comment (bam241) [reply scripts/linux_share_build.sh:28]: that would be great, but make sure you use HDF5 1.12.0, as this is the one we shipped with Cubit 2020.2

File: scripts/linux_share_build.sh (hunk 28 above)
Comment (bam241) [reply scripts/linux_share_build.sh]: I moved those echo into the yml file, so GHA was not required anymore.

my guess is using `GITHUB_ENV` or `.bashrc` is pretty similar... we still have to echo them eventually...

File: scripts/README.md [hunk 30]
```diff
@@ -16,9 +16,11 @@ Linux (tested on Ubuntu 18.04 and 20.04):
...
-    - arguments: Trelis/Cubit version to build the plugin for (only supported yet `17.1.0`)
+    - arguments: 
+        - 1 Trelis/Cubit version to build the plugin for (only supported yet `17.1.0`and Cubit `2020.2`)
+        - 2 path to Trelis/Cubit package folder (containing Trelis install package and SDK)
 - `build_plugin_docker_ubuntu.sh`: build the plugin in a docker container 
     - arguments: Arguments have to be provided in order
         - 1 docker image (tested on `ubuntu:18.04` and `ubuntu:20.04`)
         - 2 local path to the folder containing the Trelis/Cubit install deb and SDK
-        - 3 version of Cubit Trelis to build the plugin for (only tested with `17.1.0`)
+        - 3 version of Cubit Trelis to build the plugin for (only tested with `17.1.0` and Cubit `2020.2`)
```
pshriwise: Can we flip args 2 and 3 here so the order is more consistent between the docker and Linux scripts?

scripts/linux_share_build.sh
```diff
@@ -3,33 +3,68 @@ PROC=$((`grep -c processor /proc/cpuinfo`))
...
+        TRELIS_PATH="/opt/Coreform-Cubit-2020.2"
+        TRELIS_PKG="Coreform-Cubit-2020.2-Lin64.deb"
+        TRELIS_SDK_PKG="Coreform-Cubit-2020.2-Lin64-SDK.tar.gz"
+    elif [ "$1" = "17.1.0" ]; then
+        TRELIS_PATH="/opt/Trelis-17.1"
+        TRELIS_PKG="Trelis-17.1.0-Lin64.deb"
+        TRELIS_SDK_PKG="Trelis-SDK-17.1.0-Lin64.tar.gz"
+        CMAKE_ADDITIONAL_FLAGS="-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0"
+    else
+        echo "unknown Trelis/Cubit version, use: \"17.1\" or \"2020.2\""
```
pshriwise: If this is in fact an error, let's return a non-zero exit code for the function to indicate the failure and stop the script.

```suggestion
        echo "unknown Trelis/Cubit version, use: \"17.1\" or \"2020.2\""
        return 1
        
```

scripts/linux_share_build.sh
```diff
@@ -3,33 +3,70 @@ PROC=$((`grep -c processor /proc/cpuinfo`))
...
```
pshriwise:
```suggestion
        echo "unknown Trelis/Cubit version, use: \"17.1.0\" or \"2020.2\""
```

pshriwise:
Getting a docker error about the CMake version when building hdf5 w/ Ubuntu 18.04.

```
+ cmake ../hdf5 -DBUILD_SHARED_LIBS:BOOL=ON
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.12 or higher is required.  You are running version 3.10.2

```

bam241:
> Getting a docker error about the CMake version when building hdf5 w/ Ubuntu 18.04.
> 
> ```
> + cmake ../hdf5 -DBUILD_SHARED_LIBS:BOOL=ON
> CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
>   CMake 3.12 or higher is required.  You are running version 3.10.2
> ```

only tried with 20.04...

pshriwise:
From what I can see, Coreform has updated its package names for Cubit 2020.2 from `Coreform-Cubit-2020.2-Lin64.deb` to `Coreform-Cubit-2020.2-Ubuntu20.deb`. I'm wondering if it would be cleaner to require the Trelis Debian package and SDK filepaths to avoid problems like this in the future.

bam241:
I am not sure I understand what you mean...

pshriwise:
The name of the downloaded Debian package file has changed on their site:

Screenshot from 2021-03-10 22-20-39

scripts/linux_share_build.sh
```diff
@@ -56,19 +91,29 @@ function build_dagmc(){
...
     cd ../..
     rm -rf DAGMC/DAGMC DAGMC/bld
 }
 
 function setup_Trelis_sdk() {
-    cd /Trelis-sdk 
-    dpkg -i Trelis-$1-Lin64.deb
 
+    cd ${FOLDER_PKG}
+    $SUDO apt install -y ./${TRELIS_PKG}
```
pshriwise: Learning as I go here.. `apt` is meant for interactive package installs, providing colors, progress bar, etc. `apt-get` on the other hand is meant for use in scripts like one and can provide slightly different functionality than `apt`. Let's use `apt-get` in these files.

```suggestion
    $SUDO apt-get install -y ./${TRELIS_PKG}
```

bam241:
> The name of the downloaded Debian package file has changed on their site:
> 
> Screenshot from 2021-03-10 22-20-39

ok, I got that part, I was unsure about what you suggested we do about it.

My understanding is you suggest to rename the package name we have in the script.

pshriwise:
> My understanding is you suggest to rename the package name we have in the script.

My thought was that we could have two arguments, one that points to the Trelis/Cubit package file and another that points to the SDK file. We wouldn't have any hardcoded names in the scripts, making it easier to adjust to new names for these files going forward.

bam241:
this would me 4 arguments for the scripts, 

I honestly don't mind about implementing it, but wondering if it is not a lot of arguments ?

pshriwise:
> I honestly don't mind about implementing it, but wondering if it is not a lot of arguments ?

I don't find that to be too bad personally. The tradeoff in flexibility vs. verbosity seems worth it here, especially as new versions of Trelis/Cubit-Learn are released. It would be nice not to have to add blocks of code to these scripts every time there's a new version released.

bam241:
@pshriwise this should be good now.

sorry for the delay

.github/workflows/linux.yml
pshriwise: So if anything but `ubuntu:20.04` is passed with `2020.2` should the script error or print a warning?

bam241: I guess more recent version should be fine as well...
 And this isn not necessary, if one install the required deps....

anyway if the requirement are not met `apt-get` will through an error when trying to install Cubit.

I would keep that as is specially as the targeted audience of those scripts should have some experience with those tools (apt-get, docker etc...)

pshriwise: Given the target audience for these scripts, I think that's fair.

pshriwise:
I'm still getting that CMake error in hdf5 1.12.0 building in Ubuntu 18.04. If we can rely on the system hdf5 install when building the plugin on that OS maybe we should stick to that?

bam241:
@pshriwise the last commit should have fix the hdf5 issue.

basically checking the ubuntu version and installing hdf5 form apt-get when ubuntu_version <= 18

scripts/README.md
```diff
@@ -16,9 +16,15 @@ Linux (tested on Ubuntu 18.04 and 20.04):
     - `build_plugin_pkg()`: build a tarball containing the plugin
 
 - `build_plugin_linux.sh`: build the plugin on a linux computer (relies on apt-get to install required dependencies)
-    - arguments: Trelis/Cubit version to build the plugin for (only supported yet `17.1.0`)
+    - arguments: 
+        - 1 Trelis/Cubit version to build the plugin for (only supported yet `17.1.0`and Cubit `2020.2`)
+        - 2 path to Trelis/Cubit package folder (containing Trelis install package and SDK)
+        - 3 name of the Trelis deb package
+        - 4 name fo the Trelis sdk tarball
```
pshriwise:
```suggestion
        - 4 name of the Trelis sdk tarball
```

scripts/README.md
```diff
@@ -16,9 +16,15 @@ Linux (tested on Ubuntu 18.04 and 20.04):
...
 - `build_plugin_docker_ubuntu.sh`: build the plugin in a docker container 
     - arguments: Arguments have to be provided in order
         - 1 docker image (tested on `ubuntu:18.04` and `ubuntu:20.04`)
-        - 2 local path to the folder containing the Trelis/Cubit install deb and SDK
-        - 3 version of Cubit Trelis to build the plugin for (only tested with `17.1.0`)
+        - 2 version of Cubit Trelis to build the plugin for (only tested with `17.1.0` and Cubit `2020.2`)
+        - 3 local path to the folder containing the Trelis/Cubit install deb and SDK
+        - 4 name of the Trelis deb package
+        - 5 name fo the Trelis sdk tarball
```
pshriwise:
```suggestion
        - 5 name of the Trelis sdk tarball
```

scripts/build_plugin_docker_ubuntu.sh
```diff
@@ -1,4 +1,12 @@
 #!/bin/bash
 
+
+# Usage: 
+# $1: docker image to use (tested with ubuntu:18.04 and ubuntu:20.04)
+# $2: Trelis/Cubit version to install, 17.1.0 or 2020.2
+# $3 path to local pkg folder containing both trelis/cubit install package and sdk
+# $4 name of the Trelis deb package
+# $5 name fo the Trelis sdk tarball
```
pshriwise:
```suggestion
# $5 name of the Trelis sdk tarball
```

scripts/linux_share_build.sh
```diff
@@ -98,19 +148,19 @@ function build_plugin_pkg(){
...
+    patchelf --set-rpath ${TRELIS_PATH}/bin/plugins/svalinn libmakeWatertight.so
+    patchelf --set-rpath ${TRELIS_PATH}/bin/plugins/svalinn libpyne_dagmc.so
+    patchelf --set-rpath ${TRELIS_PATH}/bin/plugins/svalinn libuwuw.so
 
     # Create the Svalinn plugin tarball
     cd ..
     ln -sv svalinn/libsvalinn_plugin.so .
     cd ../..
-    tar --sort=name -czvf svalinn-plugin_linux_$1.tgz bin
+    tar --sort=name -czvf svalinn-plugin_linux_cubit_$1.tgz bin
```
pshriwise, line 164: Just noticed that at some point we'll want to set the permissions of the tarball so that users other than `root` can access it (for Docker builds).

bam241: ?
this is not build as root.
or does the docker build make is as root by default ?

bam241: Oh we run the docker as root, so it is create as root ?

bam241: I just add a line to change permission to read/write for everyone.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,127 @@
...
+            echo "CMAKE_ADDITIONAL_FLAGS=-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0" >> $GITHUB_ENV
+            echo "UBUNTU_VERSION=18" >> $GITHUB_ENV
+            echo "HDF5_PATH=/usr/lib/x86_64-linux-gnu/hdf5/serial" >> $GITHUB_ENV
+          else
+            echo "TRELIS_PKG=Coreform-Cubit-2020.2-Ubuntu20.deb" >> $GITHUB_ENV
+            echo "TRELIS_SDK_PKG=Coreform-Cubit-2020.2-Lin64-SDK.tar.gz" >> $GITHUB_ENV
+            echo "CUBIT_PATH=/opt/Coreform-Cubit-2020.2" >> $GITHUB_ENV
+            echo "UBUNTU_VERSION=20" >> $GITHUB_ENV
+            echo "HDF5_PATH=/usr/local/HDF_Group/HDF5/1.12.0" >> $GITHUB_ENV
+
```
pshriwise:
```suggestion
```

scripts/linux_share_build.sh
```diff
@@ -3,33 +3,74 @@ PROC=$((`grep -c processor /proc/cpuinfo`))
...
+    # Setup the variables
+    if [ "$1" = "2020.2" ]; then
+        TRELIS_PATH="/opt/Coreform-Cubit-2020.2"
+    elif [ "$1" = "17.1.0" ]; then
+        TRELIS_PATH="/opt/Trelis-17.1"
+        CMAKE_ADDITIONAL_FLAGS="-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0"
+    else
+        echo "unknown Trelis/Cubit version, use: \"17.1.0\" or \"2020.2\""
+        return 1
+
```
pshriwise:
```suggestion
```

pshriwise (approved):
Just added a couple of whitespace nitpicks, but I think this is looking good. Thank you for all your work @bam241. 

I'll leave this until the end of the day before merging in case anyone wants to try it out or has other comments.

bam241:
since your approval @pshriwise I added the `build` folder into `.gitignore` for convenience...

PR #85, opened 2021-02-24
Windows build compatibility -- Actions for Trelis 17.1 and Cubit 2021

bam241:
This should not be reviewed before #84 gets merged. It builds on top of #84 and relies on some of those changes.
This also relies on multiples PR to be merged on dependencies such as `MOAB`, `DAGMC` and `mncp2cad`

This includes:
- script to build Trelis plugin on Windows
- GitHub action to build the windows plugin for Trelis 17.1 and Cubit 2020.2

**This is NOT a standalone PR, it is depending on:**
- [x] svalinn/mcnp2cad#74
- [x] svalinn/dagmc#727
- [x] [MOAB/moab#524](https://bitbucket.org/fathomteam/moab/pull-requests/524)


.github/workflows/windows.yml
```diff
@@ -0,0 +1,198 @@
...
+        shell:  cmd
+        run: |
+          cd hdf
+          msiexec.exe /i HDF5-1.8.21-win64.msi /passive /norestart
+
+      - name: Build MOAB
+        shell: bash -l {0}
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          git clone https://bitbucket.org/bam241/moab -b windows_2
```
bam241: Need to be changed after `MOAB` PR gets merged.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,198 @@
...
+          cmake --build . --config Release
+          cmake --install . --config Release
+
+
+      - name: build DAGMC
+        shell: bash -l {0}
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          mkdir dagmc_build dagmc_install
+          git clone https://github.com/bam241/dagmc -b windows
```
bam241: Need to be changed after `DAGMC` PR gets merged.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,198 @@
...
+      - uses: actions/checkout@v2
+
+      - name: build plugin
+        shell: bash -l {0}
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          cp -r /d/a/Trelis-plugin/Trelis-plugin ./
+          cd Trelis-plugin
+          rm -rf mcnp2cad
+          git clone  https://github.com/bam241/mcnp2cad -b windows
```
bam241: Need to be changed after `mcnp2cad` PR gets merged.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,198 @@
...
+          cd ../
+          mkdir plugin_build plugin_install
+          cd plugin_build
+          cmake ../Trelis-plugin/ \
+                -G"Visual Studio 16 2019" \
+                -DCubit_DIR="../Trelis/bin" \
+                -DCUBIT_ROOT="../Trelis/bin" \
+                -DUPDATE_SUBMODULES=OFF \
```
bam241: Need to be changed after `mcnp2cad` PR gets merged.

gonuke:
Do we use the `build_windows.sh` script anymore?

shimwell:
Wow this will be a great addition. Super work BaM.

I notice it uses the ```run-on``` windows-2019 option. Wondering if it would also work as ```run-on``` ubuntu but use an windows-2019 ```image``` for the build then we could perhaps easily include other images like the windows-2022 image if needed

bam241:
> Do we use the `build_windows.sh` script anymore?

we never had one for windows.

gonuke:
> > Do we use the `build_windows.sh` script anymore?
> 
> 
> 
> we never had one for windows.

But I don't see it being used anywhere because the GH action explicitly includes all the build instructions instead

bam241:
> > > Do we use the `build_windows.sh` script anymore?
> > 
> > 
> > we never had one for windows.
> 
> But I don't see it being used anywhere because the GH action explicitly includes all the build instructions instead

sorry, I lost track of that file..

this is an artefact of previous work. that will need to be removed eventually...

bam241:
> Wow this will be a great addition. Super work BaM.
> 
> I notice it uses the `run-on` windows-2019 option. Wondering if it would also work as `run-on` ubuntu but use an windows-2019 `image` for the build then we could perhaps easily include other images like the windows-2022 image if needed

yes this is something we should/could consider in the future. I would like to keep this PR simple, as a first support of windows.  I think we would need first to update that on the DAGMC CI.... to extend our windows capability. 
One should also check that multiple windows build would be required, maybe (if we are lucky) a single build would be compatible with multiple windows version.... (as for Mac OS)

bam241:
current problem seems to be on core form side...
https://coreform.com/products/downloads/

don't see any download available on their website.... maybe their server crashed....

shimwell:
> current problem seems to be on core form side...
> https://coreform.com/products/downloads/
> 
> don't see any download available on their website.... maybe their server crashed....

I think it might be the cloud host https://downdetector.com/status/backblaze/

.github/workflows/windows.yml
```diff
@@ -0,0 +1,237 @@
...
+    paths-ignore:
+      - '.github/workflows/unix_linux.yml'
+      - '.github/workflows/unix_mac.yml'
+      - 'scripts/**'
+      - '*.md'
+      - 'LICENSE'
+
+  release:
+    types: # This configuration does not affect the page_build event above
+      - created
```
gonuke: I think we learned last week that `created` was not a good choice. It is possible to create a draft release, which then triggers this, but when you publish the release, it does not trigger again.


@Shimwell , did we pick a better one? Or maybe an additional one?

gonuke (changes requested):
We need to choose best release actions to trigger a build

.github/workflows/windows.yml
shimwell: I believe we preferred ```published```

shimwell: All the options are here https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release

There is also a discussion here https://github.community/t/what-is-the-created-type-in-on-release/17184

bam241:
@gonuke this looks good to go ? 
Anya additional comments/requests ?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,229 @@
...
+      - name: HDF5 download
+        shell:  bash -l {0}
+        run: |
+          curl -L https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.8/hdf5-1.8.21/bin/hdf5-1.8.21-Std-win7_64-vs14.zip --output hdf5-1.8.21-Std-win7_64-vs14.zip
+          unzip hdf5-1.8.21-Std-win7_64-vs14.zip
+      - name: HDF5 install
+        shell:  cmd
+        run: |
+          cd hdf
+          msiexec.exe /i HDF5-1.8.21-win64.msi /passive /norestart
```
gonuke: question: does this mean that conda's HDF5 won't work? also the conda-forge version?  I wonder if we can document this somewhere so that if/when someone tries again, they know what the problem(s) were

.github/workflows/windows.yml
```diff
@@ -0,0 +1,229 @@
...
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          mkdir -p bin/plugins
+          cd bin/plugins
+          cp ../../moab_install/bin/MOAB.dll ./
+          cp ../../dagmc_install/lib/*.lib ./
+          cp ../../plugin_install/bin/* ./
+          cp ../../plugin_install/lib/* ./
+          cd ${PLUGIN_ABS_PATH}
+          tar -cvf svalinn_plugin_windows_${{ matrix.cubit }}.tar  bin
```
gonuke: should we use `7z` to make a zip file instead of a tarball?

gonuke (changes requested):
Thanks @bam241 - exciting to see this getting close.  Two items, only one is blocking.

bam241:
@gonuke it seems to be working, using hdf5 conda and 7x :)

.github/workflows/windows.yml
```diff
@@ -0,0 +1,213 @@
...
+        run: |
+          cd ${PLUGIN_ABS_PATH}
+          mkdir -p bin/plugins
+          cd bin/plugins
+          cp ../../moab_install/bin/MOAB.dll ./
+          cp ../../dagmc_install/lib/*.lib ./
+          cp ../../plugin_install/bin/* ./
+          cp ../../plugin_install/lib/* ./
+          cd ${PLUGIN_ABS_PATH}
+          7z a C:/Users/runneradmin/svalinn_plugin_windows_${{ matrix.cubit }}.7z  bin\
```
gonuke: Can we make this with a `.zip` extension and ZIP format?

gonuke (changes requested):
Can we get the ZIP format?

.github/workflows/windows.yml
gonuke: P.S. Thanks for figuring the rest out, @bam241 !

gonuke (approved):
Excellent!!! Thanks @bam241

PR #73, opened 2021-01-28
Windows build

bam241:
this is a start to build plug on windows.

tested for Trelis 17.1

this is producing a archive that contain all the file required for a plugin (just dump all the file in the plugin folder.)

I did not do extended testing on Trelis just exporting a empty geometry...

PR #86, opened 2021-02-24
MacOS build script & Github Action for Trelis 17.1 and Cubit 2020.2

bam241:
This builds on top of #84 , review after merge.
This also relies on svalinn/mcnp2cad#73

This includes:
- build script to build the plugin on MacOS
- GItHub Action to build the plugin on MacOS for Trelis17.1 and Cubit 2020.2


.github/workflows/mac.yml
```diff
@@ -0,0 +1,182 @@
...
+          make install
+          cd ../..
+          rm -rf DAGMC/DAGMC DAGCM/bld
+
+      - name: Build plugin
+        shell: bash
+        run: |
+          cd ${PLUGIN_ABS_PATH}/Trelis-plugin
+          rm -rf mcnp2cad
+          git clone  https://github.com/bam241/mcnp2cad -b mac_build
```
bam241: this needs to be changed when svalinn/mcnp2cad#73, gets merged.

.github/workflows/mac.yml
```diff
@@ -0,0 +1,182 @@
...
+
+          mkdir -pv bld
+          cd bld
+          cmake .. $CMAKE_ADDITIONAL_FLAG \
+                  -DUPDATE_SUBMODULES=OFF \
```
bam241: this needs to be changed when svalinn/mcnp2cad#73, gets merged.

bam241:
this works hdf5@1.12 added a new PR depedency: svalinn/dagmc#728

we could avoid svalinn/dagmc#728 and limit hdf5 version to `1.10.7`

bam241:
getting the following error when trying to load Cubit2020.2 plugin built with the GitHub action:
```
ERROR: <stdin>, line 0. Error loading /Applications/Coreform-Cubit-2020.2.app/Contents/MacOS/plugins/libsvalinn_plugin.so:
ERROR: <stdin>, line 0. 
Finished Command: undo on
```

bam241:
installing `brew install eigen hdf5` using brew solves it.

I guess we have to ship some of those within the package

bam241:
it was actually slightly more complicated than that:
one of the `hdf5` deps was required not` hdf5` itself `brew install szip`

bam241:
I fixed it adding `/libsz.2.dylib` in the plugin and running :
`sudo install_name_tool -change /usr/local/opt/szip/lib/libsz.2.dylib @rpath//libsz.2.dylib libsvalinn_plugin.so`

bam241:
inclusion of szlib fixed it.

gonuke:
I plan to review this after we merge #84 to focus on the Mac parts...

bam241:
Reworking this so it can match the linux design, 
It'll be broken for a while, please ignore this until I ping you :)

bam241:
@gonuke @pshriwise I reshaped the MacOS build to leverage the linux_share_build.sh (now named unix_shared_build.sh)
I have no way right now to test the `build_plugin_macos.sh`.

I didn't re-tried the linux docker build against those new change as well.

.github/workflows/mac.yml
```diff
@@ -0,0 +1,122 @@
...
+          fi
+          echo "BUILD_SHARED_LIBS=OFF" >> $GITHUB_ENV
+          echo "BUILD_STATIC_LIBS=ON" >> $GITHUB_ENV
+          echo "source $GITHUB_WORKSPACE/scripts/unix_share_build.sh" >> $HOME/.profile
+
+
+      - name: Initial setup
+        shell: bash -l {0}
+        run: |
+          source $HOME/.profile
```
gonuke: It is weird that this doesn't get sourced automatically...  I wonder if it is sourcing something else? (e.g. possibly .bash_profile per bash documentation?)  Alternatively, is there a way we can launch bash in the `shell` command to ensure it is sourced?

.github/workflows/mac.yml
```diff
@@ -0,0 +1,122 @@
+name: MACOS
```
gonuke, line 1: In general, I wonder if we can nudge this to be even more similar to the linux action, just for long term maintenance purposes

scripts/CubitExport-release.cmake
```diff
@@ -0,0 +1,190 @@
+#----------------------------------------------------------------
```
gonuke: Do we need this file? or is it part of the SDK?

scripts/CubitGeomConfig.cmake
```diff
@@ -0,0 +1,116 @@
+# Generated by CMake
```
gonuke: Do we need this file in the repo?

scripts/build_plugin_linux.sh
```diff
@@ -11,27 +11,29 @@ FOLDER_PKG="$2"
...
+linux_setup_var $1
 
 build_hdf5
 
 build_moab
 build_dagmc
 
-# $1 is the version of Trelis/Cubit one are trying to compile against i.e. 17.1.0
-setup_Trelis_sdk $1
+# $1 is the version of CUBIT/Cubit one are trying to compile against i.e. 17.1.0
```
gonuke: Should probably leave the reference to Trelis here as an option since it 17.1.0 is still called Trelis.
```suggestion
# $1 is the version of Trelis/Cubit one are trying to compile against i.e. 17.1.0
```

scripts/build_plugin_linux.sh
```diff
@@ -11,27 +11,29 @@ FOLDER_PKG="$2"
...
+linux_setup_Trelis_sdk $1
```
gonuke: It looks like we're largely changing things to refer to Cubit instead of Trelis - should we change this function name, as well?

scripts/build_plugin_macos.sh
```diff
@@ -0,0 +1,117 @@
...
+
+cd ${PLUGIN_ABS_PATH}
+ln -s $SCRIPTPATH/ ./
+
+mkdir -pv moab/bld
+cd moab
+git clone https://bitbucket.org/fathomteam/moab -b Version5.1.0
+cd bld
+cmake ../moab -DENABLE_HDF5=ON \
+          -DCMAKE_PREFIX_PATH=/usr/lib/x86_64-linux-gnu/hdf5/serial \
```
gonuke: This doesn't look right...

scripts/build_plugin_macos.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
```
gonuke: This doesn't look like its using anything out of `unix_share_build.sh` yet?

gonuke (changes requested):
Maybe your comment above answers one of my questions - but it doesn't look like `build_plugin_macos.sh` uses any of the `unix_shared_build.sh` yet...

.github/workflows/mac.yml
bam241: sadly the bash shell is run internally with the --no-profile argument which prevent on macOS to source profile/bash_profile by default :(
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell
we don't have the same behavior with linux...

scripts/CubitExport-release.cmake
bam241: yes it is part of the sdk, I had to tweak it to have it work on Mac, 
some paths were not provided correctly...

scripts/build_plugin_linux.sh
bam241: I think this is the way forward...

scripts/build_plugin_macos.sh
bam241: no I did not touch this one yet, as I don't have access to the MACOS vm.

scripts/CubitGeomConfig.cmake
bam241: I don't quite remember precisely, but it was part of the file I had to modify to have the plugin make to work.

I can do some test to check if this is still necessary...

scripts/build_plugin_linux.sh
bam241: I did a Trelis to CUBIT replacement for ENV variable, it caught to much stuff... need to revert some of them...

.github/workflows/mac.yml
gonuke: odd - it says that Linux should behave the same....

scripts/CubitExport-release.cmake
gonuke: I wonder if we should provide that differently then???  It seems to make more sense to patch the SDK that include the whole file here, but I'm not sure the best way to do that....

bam241:
our PyNE amalgamation broke this...

scripts/CubitExport-release.cmake
bam241: we could get a diff, and apply it ?

.github/workflows/mac.yml
gonuke: Actually... our declaration of the shell should override this on both platforms

I wonder if there is also a `.bash_profile` that is being read instead when we use `-l`?

scripts/build_plugin_macos.sh
gonuke: So what's the next step for this file?

bam241: maybe I can remove it from this. 
and PR it when I am back ?

.github/workflows/mac.yml
```diff
@@ -0,0 +1,122 @@
...
+        run: |
+          echo "CURRENT=$(pwd)" >> $GITHUB_ENV
+          echo "SCRIPTPATH=$GITHUB_WORKSPACE" >> $GITHUB_ENV
+          echo "PLUGIN_ABS_PATH=$GITHUB_WORKSPACE/.." >> $GITHUB_ENV
+          echo "FOLDER_PKG=$GITHUB_WORKSPACE/../pkg" >> $GITHUB_ENV
+          echo "CUBIT_PATH=/Applications/${{ matrix.CUBIT_BASE_NAME }}.app/Contents" >> $GITHUB_ENV
+          if [ "${{ matrix.cubit }}" = "17.1.0" ]; then
+            echo "CMAKE_ADDITIONAL_FLAGS=-DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0" >> $GITHUB_ENV
+            echo "TRELIS_PKG=Trelis-17.1.0-Mac64.dmg" >> $GITHUB_ENV
+            echo "TRELIS_SDK_PKG=Trelis-SDK-17.1.0-Mac64.tar.gz" >> $GITHUB_ENV
```
gonuke: Should these be changed to CUBIT as well?
```suggestion
            echo "CUBIT_PKG=Trelis-17.1.0-Mac64.dmg" >> $GITHUB_ENV
            echo "CUBIT_SDK_PKG=Trelis-SDK-17.1.0-Mac64.tar.gz" >> $GITHUB_ENV
```

gonuke (changes requested):
Some lingering references to trelis in variable names

gonuke:
Still curious about making the `mac.yml` file a little more similar to the `linux.yml` file, but that might be best as a conversation

gonuke:
I PR'ed some changes to address

gonuke (approved):
Given my contribution, I probably shouldn't merge this.... but I approve

gonuke:
I think this is ready to go @pshriwise

bam241:
as a reminder this depends on svalinn/mcnp2cad#73

bam241:
@pshriwise when you have time could you take a look here ?

.github/workflows/mac.yml
pshriwise: Seconded. I wonder if a  job parameter indicating the OS type and some [conditionals](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsif) in the job steps that check that parameter 
would do the trick.

scripts/CubitExport-release.cmake
pshriwise: That sounds like a good plan to me @bam241!

scripts/CubitGeomConfig.cmake
pshriwise: Maybe we can add this to the list of files to get diffs of for a patch.

scripts/unix_share_build.sh
```diff
@@ -66,13 +95,12 @@ function build_moab() {
...
     cmake ../moab -DENABLE_HDF5=ON \
             -DHDF5_ROOT=$HDF5_PATH \
-            -DBUILD_SHARED_LIBS=ON \
+            -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} \
             -DENABLE_BLASLAPACK=OFF \
             -DENABLE_FORTRAN=OFF \
             $CMAKE_ADDITIONAL_FLAGS \
             -DCMAKE_INSTALL_PREFIX=${PLUGIN_ABS_PATH}/moab
-
-    make -j$PROC
```
pshriwise, line 75: I'm curious about this. Why did it need to change? Should we select only a few processes instead?

pshriwise (changes requested):
Overall this looks pretty good to me. There are just some outstanding issues that @gonuke raised I think:

  - looking into merging the Linux and Mac workflow files
  - getting diff's of the Cubit CMake files from the SDK and look into patching those instead of including the modified files

Also if we can communicate the changes we needed to make to the CMake files Coreform that would probably be good.

scripts/unix_share_build.sh
bam241: I removed it because we are not here for speed, and this something causes problems with `gcc`....

scripts/build_plugin_linux.sh
```diff
@@ -11,27 +11,29 @@ FOLDER_PKG="$2"
...
 CMAKE_ADDITIONAL_FLAGS=""
 UBUNTU_VERSION=""
 HDF5_PATH=""
+BUILD_SHARED_LIBS=""
 
-source ${SCRIPTPATH}/scripts/linux_share_build.sh
+
+source ${SCRIPTPATH}/scripts/unix_share_build.sh
 
 install_prerequisites
```
bam241:
```suggestion
linux_install_prerequisites
```

bam241:
@pshriwise @gonuke I replaced the cmake files with a .diff files and apply a patch now instead of replacing those 2 files...

I think this is ready to go in
(as we said the merge between Mac.yml and linux.yml should go in a later PR)

scripts/build_plugin_linux.sh
```diff
@@ -11,27 +11,29 @@ FOLDER_PKG="$2"
...
 build_hdf5
 
 build_moab
 build_dagmc
 
 # $1 is the version of Trelis/Cubit one are trying to compile against i.e. 17.1.0
-setup_Trelis_sdk $1
+linux_setup_cubit_sdk $1
 build_plugin $1
 build_plugin_pkg $1
```
bam241:
```suggestion
linux_build_plugin_pkg $1
```

scripts/unix_share_build.sh
```diff
@@ -89,29 +117,73 @@ function build_dagmc(){
...
+        rm -rf cubit.pkg
+    fi
+
+    cd ${CUBIT_PATH}
+    if [ "${1}" = "2020.2" ]; then
+        CUBIT_BASE_NAME="Coreform-Cubit-2020.2"
+    elif [ "${1}" = "17.1.0" ]; then
+        CUBIT_BASE_NAME="Trelis-17.1"
+    fi
+    sudo tar -xzf ${FOLDER_PKG}/${CUBIT_SDK_PKG}
```
pshriwise, line 153: Should these be `$SUDO` instead of `sudo`? I can't recall.

scripts/unix_share_build.sh
```diff
@@ -89,29 +117,73 @@ function build_dagmc(){
...
     make install
     cd ../..
     rm -rf DAGMC/DAGMC DAGMC/bld
 }
 
-function setup_Trelis_sdk() {
+function mac_setup_cubit_sdk() {
+    cd ${FOLDER_PKG}
+    if [ "${1}" = "17.1.0" ]; then
+        hdiutil convert ${CUBIT_PKG} -format UDTO -o trelis_eula.dmg.cdr
```
pshriwise, line 137: Gonna request a comment or two here on the different setup methods and commands used.

CMakeLists.txt
```diff
@@ -46,6 +46,14 @@ message(STATUS "CMAKE_INSTALL_RPATH: ${CMAKE_INSTALL_RPATH}")
 set(PLUGIN_SRC_FILES SvalinnPlugin.cpp)
 set(PLUGIN_LINK_LIBS ${CUBIT_LIBS})
 
+# HDF5 for mac
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+  SET(HDF5_USE_STATIC_LIBRARIES ON)
```
pshriwise, line 51: Does use of HDF5 static libraries have any implications for how we build our DAGMC or plugin libraries?

scripts/unix_share_build.sh
```diff
@@ -89,29 +117,73 @@ function build_dagmc(){
...
+    sudo cp -pv ${CUBIT_PATH}/MacOS/CubitExport-release.cmake ${CUBIT_PATH}/MacOS/CubitExport-release.cmake.orig
+    sudo gsed -i "s/\/${CUBIT_BASE_NAME}.app\/Contents//" ${CUBIT_PATH}/MacOS/CubitExport-release.cmake
+
+    sudo cp -pv ${CUBIT_PATH}/MacOS/CCubitGeomConfig.cmake ${CUBIT_PATH}/MacOS/CubitGeomConfig.cmake.orig
+    sudo gsed -i "s/\${_IMPORT_PREFIX}\/include/\${_IMPORT_PREFIX}\/${CUBIT_BASE_NAME}.app\/include/" ${CUBIT_PATH}/MacOS/CubitGeomConfig.cmake
+
+    if [ "${1}" = "2020.2" ]; then
+        cd ${CUBIT_PATH}/bin
+        sudo cp -pv CubitExport.cmake CubitExport.cmake.orig
+        sudo gsed -i "s/\"\/\.\.\/app_logger\"/\"\"/" CubitExport.cmake
```
pshriwise, line 174: Some comments for what we're doing here would be nice also.

scripts/unix_share_build.sh
bam241: I used `sudo` for Mac instead of `$SUDO` because it should be required regardless of the platform you are using (docker, CI, local) where on ubuntu `sudo` is not required on docker as we run as root...

CMakeLists.txt
bam241: one need to use static HDF5 to allow o be build the plugin as a single `.dylib`

Or at least this is what I understood from my digging few months ago....

scripts/unix_share_build.sh
pshriwise: Ah, nice. Didn't know that you don't get root access in an OSX docker image.

pshriwise (approved):
Thanks @bam241! On to the next PR I suppose 😃

PR #89, opened 2021-03-23
Fix MACRO that checks for MOAB error

### LINKED ISSUE #88: bug: macro function in DAGMCExportCommand.cpp
```
#define CHK_MB_ERR_RET_MB(A,B)  if (moab::MB_SUCCESS != (B)) { \
  message << (A) << (B) << std::endl;                                   \
  return rval;                                                         \
  }
```

Wont return A or B instead of rval, which is not in the parameter list? 

Although, it does not report error during compiling.

gonuke:
Fixes #88 

This never caused an error in practice, but was incorrect.


bam241:
this looks good to me, we should not worry about the GitHub action failing.
this is coming form elsewhere.

bam241 (dismissed):

pshriwise:
Pretty simple change here. Hopefully it just needs to be rebased to get tests passing?

PR #90, opened 2021-03-23
Updating for PyNE Submodule in DAGMC

gonuke:
Pushing this change in ahead of #86 for more immediate effect

### DIFF
--- a/scripts/linux_share_build.sh
+++ b/scripts/linux_share_build.sh
@@ -83,7 +83,9 @@ function build_dagmc(){
     mkdir -pv DAGMC/bld
     cd DAGMC
     git clone https://github.com/svalinn/DAGMC -b develop
-    cd bld
+    cd DAGMC
+    git submodule update --init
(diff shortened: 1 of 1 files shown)


gonuke:
Do you know what's wrong here, @bam241 ?

pshriwise:
I'm guessing there are a good number of installation scripts out there that will run into this same trouble. I'm thinking it would be good to have DAGMC initialize the PyNE repo as part of the CMake step. I'd be happy to put in that PR (into DAGMC).

bam241:
there is 2 issues there:
- the secret are not accessible for PR issued from a fork.
- it is now pulling the new DAGMC with amalgamated PyNE.

the first one, when I designed the GitHub action I pictured it as run on merge, so secret would be accessible not an PR...
for the second one a fix is included in #86 

bottom line, I'd probably merge that one.

gonuke:
Closing in favor of #91

PR #91, opened 2021-03-23
Updating for PyNE Submodule in DAGMC

gonuke:
This replaces #90 with the hopes that this overcomes the lack of access to the necessary secrets, and still gets this quick fix in ahead of #86.

### DIFF
--- a/scripts/linux_share_build.sh
+++ b/scripts/linux_share_build.sh
@@ -83,7 +83,9 @@ function build_dagmc(){
     mkdir -pv DAGMC/bld
     cd DAGMC
     git clone https://github.com/svalinn/DAGMC -b develop
-    cd bld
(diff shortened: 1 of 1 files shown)


gonuke:
Noting that on #90 , @pshriwise recommended that the submodule update happen as part of the DAGMC build, perhaps via CMake

bam241:
I have a similar fix included in #85.

This looks good to me.

Shall we merge that as a temporary fix ?

bam241:
I agree that the update of the submodule in cmake is probably cleaner.

gonuke:
I'm withdrawing this in favor of other recent changes

pshriwise:
@gonuke are you still ok with closing this?

PR #92, opened 2021-03-23
Attempt to streamline action files

gonuke:
Testing some changes in the action scripts


gonuke:
This did not appear to trigger actions - will try to launch again

PR #93, opened 2021-03-23
Streamlining linux & mac actions to be similar

gonuke:
Here is my attempt to make the mac & Linux a little more similar.

Also figured out the profile issue...


bam241:
Thx a ton @gonuke.
What was the problem with the .bash_profile ?!

PR #97, opened 2021-05-24
updated the build script

makeclean:
There are probably bigger sustainability questions about how we deal with the new micro releases, but this is what I did to make it work.


scripts/linux_share_build.sh
```diff
@@ -1,13 +1,25 @@
 #!/bin/bash
+set -ue
+
+SUDO=""
```
bam241: this should be moved in the `scripts/build_plugin_linux.sh` 
otherwise It overwrites the variable for CI

bam241:
I have only 1 comments to enable GitHub action to work.

have you tried it for 2021.3 as well ?

thx a lot @makeclean

MigMagan:
A couple of changes that seem to be needed...

OS name detection is not working here
```diff --git a/scripts/unix_share_build.sh b/scripts/unix_share_build.sh
index d23b327..f7c714d 100644
--- a/scripts/unix_share_build.sh
+++ b/scripts/unix_share_build.sh
@@ -55,11 +55,10 @@ function setup_var() {
         echo "unknown Trelis/Cubit version"
         return 1
     fi
-
-    if [ "OS" == "MAC" ]; then
+    if [ $OS == "MAC" ]; then
         BUILD_SHARED_LIBS="OFF"
         BUILD_STATIC_LIBS="ON"
-    elif [ "OS" == "UBUNTU"]; then
+    elif [ $OS == "UBUNTU" ]; then
         BUILD_SHARED_LIBS="ON"
         BUILD_STATIC_LIBS="OFF"
``` 
Name of the function seems outdated
```diff --git a/scripts/build_plugin_linux.sh b/scripts/build_plugin_linux.sh
index 55d68bc..2467b17 100755
--- a/scripts/build_plugin_linux.sh
+++ b/scripts/build_plugin_linux.sh
@@ -25,19 +25,23 @@ SED="sed"
 source ${SCRIPTPATH}/scripts/unix_share_build.sh

-build_hdf5
+linux_build_hdf5
```

Also, I noticed that the plugin builds againts `libMOAB.so.0`, while the lib file is named `libMOAB.so.5`. I just linked it, and after that, the plugin is working with 2021.3 !

gonuke:
Do we need all three of these workflow files? or does `unix` replace `linux` and `macos`?

bam241:
> Do we need all three of these workflow files? or does `unix` replace `linux` and `macos`?

no I kept them to help me unifying them.

we will only keep the `unix.yml` one

bam241:
> A couple of changes that seem to be needed...
> 
> OS name detection is not working here
> 
> ```diff
> index d23b327..f7c714d 100644
> --- a/scripts/unix_share_build.sh
> +++ b/scripts/unix_share_build.sh
> @@ -55,11 +55,10 @@ function setup_var() {
>          echo "unknown Trelis/Cubit version"
>          return 1
>      fi
> -
> -    if [ "OS" == "MAC" ]; then
> +    if [ $OS == "MAC" ]; then
>          BUILD_SHARED_LIBS="OFF"
>          BUILD_STATIC_LIBS="ON"
> -    elif [ "OS" == "UBUNTU"]; then
> +    elif [ $OS == "UBUNTU" ]; then
>          BUILD_SHARED_LIBS="ON"
>          BUILD_STATIC_LIBS="OFF"
> ```
> 
> Name of the function seems outdated
> 
> ```diff
> index 55d68bc..2467b17 100755
> --- a/scripts/build_plugin_linux.sh
> +++ b/scripts/build_plugin_linux.sh
> @@ -25,19 +25,23 @@ SED="sed"
>  source ${SCRIPTPATH}/scripts/unix_share_build.sh
> 
> -build_hdf5
> +linux_build_hdf5
> ```
> 
> Also, I noticed that the plugin builds againts `libMOAB.so.0`, while the lib file is named `libMOAB.so.5`. I just linked it, and after that, the plugin is working with 2021.3 !

@MigMagan thx for your feedback.

the OS is not supposed to be a detection but a variable set by the user (see the updated comments), also I haven't try to script yet, I was focussing on the GitHub action build...

about the libMOAB.so, I am unsure about what you mean by linked it, but it should be included in the plugin folder as we import `${PLUGIN_ABS_PATH}/moab/lib/libMOAB.so*`...

bam241:
@makeclean @gonuke @pshriwise this should be ready for a extensive review.

I also included a merge of the linux and Mac GitHub action into unix GitHub action.

I only included the 17.1.0 and 2021.4 version yet as this is the only version we have both .deb for ubuntu and .dmg for Mac ...

note that 2021.5 has just been released.....

scripts/unix_share_build.sh
```diff
@@ -31,44 +53,41 @@ function setup() {
...
+        BUILD_STATIC_LIBS="OFF"
+    else
+        echo "OS ENV variable needs to be defined to either UBUNTU or MAC"
+        return 1
+    fi
 }
 
-function build_hdf5() {
+function linux_build_hdf5() {
     # if ubuntu 18.04 or lower rely of apt-get hdf5
```
pshriwise:
```suggestion
    # if ubuntu 18.04 or lower rely on apt-get hdf5
```

scripts/unix_share_build.sh
```diff
@@ -131,66 +154,66 @@ function build_dagmc(){
...
+    if [ "$1" == "2021.3" ] || [ "$1" == "2021.4" ] ; then
+	    return
+    fi
+
     cd /opt
     $SUDO tar -xzf ${FOLDER_PKG}/${CUBIT_SDK_PKG}
     # removing app_loger that seems to not be present in Cubit 2020.2
     if [ "$1" = "2020.2" ]; then
         cd ${CUBIT_PATH}/bin
         $SUDO cp -pv CubitExport.cmake CubitExport.cmake.orig
```
pshriwise: The lines inside the if statement are identical to the ones in the `mac_setup_cubit` function. IMO it would be good to separate these commands into their own function so we don't miss an update to the regex in one or the other.

scripts/unix_share_build.sh
```diff
@@ -201,7 +224,7 @@ function build_plugin(){
     cd ../
     mkdir -pv bld
     cd bld
-    cmake ../Trelis-plugin -DCUBIT_ROOT=${CUBIT_PATH} \
+    cmake ../Trelis-plugin -DCMAKE_PREFIX_PATH=${CUBIT_PATH} -DCUBIT_ROOT=${CUBIT_PATH} \
```
pshriwise:
```suggestion
    cmake ../Trelis-plugin -DCMAKE_PREFIX_PATH=${CUBIT_PATH} \
                           -DCUBIT_ROOT=${CUBIT_PATH} \
```

pshriwise (changes requested):
Just a few small comments/suggestions. Nice to have only one YAML to worry about now!

scripts/unix_share_build.sh
bam241: done

thx for the review !

pshriwise (approved):
This is looking good to me. Onward! Thanks @makeclean @bam241!

PR #101, opened 2021-06-15
Attempt to publish binary on release

bam241:
this is an attempt to publish out build on release.

not sure how that will work tho.... or how to test it....


bam241:
this should be working now and produce something like: https://github.com/bam241/Trelis-plugin/releases/tag/0.0.8
(upload to come as CI finishes)

.github/workflows/unix.yml
```diff
@@ -132,10 +134,18 @@ jobs:
...
+          path: ~/svalinn-plugin_${{ matrix.os }}_cubit_${{ matrix.cubit }}.tgz
+    
+      - if: github.event_name == 'release'
+        name: Upload binaries to release
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: ~/svalinn-plugin_${{ matrix.os }}_cubit_${{ matrix.cubit }}.tgz
+          asset_name: svalinn-plugin_${{ matrix.os }}_cubit_${{ matrix.cubit }}.tgz
+          tag: ${{ github.ref }}
```
gonuke: add a newline

.github/workflows/unix.yml
```diff
@@ -132,10 +134,18 @@ jobs:
         run: |
           ${system}_build_plugin_pkg ${{ matrix.cubit }}
 
-      - name: Upload package
+      - if: github.event_name != 'release'
+        name: Upload package
```
gonuke, line 139: where does this get uploaded?

gonuke (changes requested):
Thanks @bam241 - looking forward to getting this working!

.github/workflows/unix.yml
bam241: should have been upload there: https://github.com/bam241/Trelis-plugin/releases/tag/0.0.8

but there is a problem....

bam241: basically we should end up with something like:
https://github.com/bam241/Trelis-plugin/releases/tag/0.0.7

bam241:
this should be ready,
you can see an example of use on my fork:
item get added automatically here: https://github.com/bam241/Trelis-plugin/releases/tag/0.0.12
when each job on CI get done they will be added in the release page
(corresponding jobs: https://github.com/bam241/Trelis-plugin/actions/runs/943758101)

bam241:
cc @gonuke

bam241:
the action I added automatically runs on new release, and add the artefact as binaries in the release, (where the release has been created)

I ran it on my fork to test it and avoid spamming and polluting `develop` here, but this is not kink to a fork, will add the artifacts where the release has been issued

.github/workflows/unix.yml
pshriwise: Wait I'm confused too, this block gets executed when the event is not a release right? Do we want to upload anything in that case?

bam241: this is upload the artifact at the job summary location they're are useful for debug when running CI, 
on release we use the other upload mechanism

pshriwise: Ahhh I see. Very nice! Didn't know that was a thing.

pshriwise (approved):
Great addition @bam241! So nice to have these automatically published in a place that's easily accessible for new users.

gonuke (approved):
LGTM - thanks @bam241

PR #107, opened 2021-07-23
Forcing use of python3 in CI

bam241:
This should enforce python3 in CI and should comply with new DAGMC requirement

### DIFF
--- a/scripts/unix_share_build.sh
+++ b/scripts/unix_share_build.sh
@@ -134,7 +134,7 @@ function build_dagmc(){
     cd ${PLUGIN_ABS_PATH}
     mkdir -pv DAGMC/bld
     cd DAGMC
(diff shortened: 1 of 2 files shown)


scripts/unix_share_build.sh
```diff
@@ -134,7 +134,7 @@ function build_dagmc(){
     cd ${PLUGIN_ABS_PATH}
     mkdir -pv DAGMC/bld
     cd DAGMC
-    git clone https://github.com/svalinn/DAGMC -b develop
+    git clone https://github.com/bam241/DAGMC -b py3
```
bam241, line 137: this needs to change

bam241:
this is nor relevant anymore, it has been fixed on the dagmc side.
closing

PR #110, opened 2021-07-29
no bin folder for mac?

bam241:
trying to fix the missing bin folder for trelis17.1

### DIFF
--- a/scripts/unix_share_build.sh
+++ b/scripts/unix_share_build.sh
@@ -279,9 +279,9 @@ function mac_build_plugin_pkg(){
 
     # restoring correct RPATH for 17.1 (bin does not exist as it is not shipped with SDK)
     if [ "$1" == "17.1.0" ] ; then
-        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/nn libsvalinn_plugin.so
-        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/nn libiGeom.dylib
-        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/nn libmcnp2cad.dylib
+        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/svalinn libsvalinn_plugin.so
+        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/svalinn libiGeom.dylib
+        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/svalinn libmcnp2cad.dylib
         install_name_tool -rpath ${CUBIT_PATH}/bin ${CUBIT_PATH}/MacOS libmcnp2cad.dylib
         install_name_tool -rpath ${CUBIT_PATH}/bin ${CUBIT_PATH}/MacOS libiGeom.dylib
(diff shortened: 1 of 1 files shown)


bam241:
I tested it locally, plugin is now loading for MacOS 17.1

bam241:
@lewisgross1296
have you been able to test the plugin from there ?

lewisgross1296:
After re-installing Trelis-17.1, patching it, and moving the un-tar'd [plugin](https://github.com/svalinn/Cubit-plugin/suites/3364392662/artifacts/79152110) to `/Applications/Trelis-17.1.app/Contents/MacOS/plugins`, I was able to open Trelis and view the message

```Loaded Svalinn plugin.
-- DAGMC export command available.
-- iGeom_test command available.
-- MCNP import command available.
Finished Command: set node constraint smart
Finished Command: undo on
```

After this, I entered `help dagmc` and it output the help information

`export dagmc <filename> [faceting_tolerance <faceting tolerance>] [length_tolerance <length tolerance>]
     [normal_tolerance <normal tolerance>] [make_watertight] [verbose]
     [fatal_on_curves]`

gonuke:
What "patch" are you applying to Trelis 17.1?

bam241:
basically, when the SDK is present, there is a `bin` symbolic link that point to `MacOS`.

the RPATH the lib uses `bin` and not `MacOS`. What I do in this is reset the RPATH to use `MacOS` instead of `bin`. it is only required for 17.1 because that the only version we support that does not ship with the SDK (so with the `bin` symbolic link)

lewisgross1296:
@bam sent me a patch on box that I had to unpack and move to `/Applications/Trelis-17.1.app/Contents/MacOS`. It was discussed in #svalinn-support on slack on Monday July 26. I think without it I couldn't log into the cnerg license server, so it probably isn't something that everyone needs to patch, just people using our license server.

bam241:
> What "patch" are you applying to Trelis 17.1?

I think he is refering the lib patch required to use an updated license server, this has nothing to do with the plugin

scripts/unix_share_build.sh
```diff
@@ -277,6 +277,17 @@ function mac_build_plugin_pkg(){
     cp /usr/local/opt/szip/lib/libsz.2.dylib .
     install_name_tool -change /usr/local/opt/szip/lib/libsz.2.dylib @rpath/libsz.2.dylib libsvalinn_plugin.so
 
+    # restoring correct RPATH for 17.1 (bin does not exist as it is not shipped with SDK)
+    if [ "$1" == "17.1.0" ] ; then
+        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/svalinn libsvalinn_plugin.so
+        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/svalinn libiGeom.dylib
+        install_name_tool -rpath ${CUBIT_PATH}/bin/plugins/svalinn ${CUBIT_PATH}/MacOS/plugins/svalinn libmcnp2cad.dylib
```
gonuke, line 284: Do these put two paths in the path? Do we need both, or just not know which one?

gonuke (commented):
I think this is good - just want to understand what this change does

scripts/unix_share_build.sh
bam241: this is replacing the first one by the second:
```
INSTALL_NAME_TOOL(1)                                                                                      INSTALL_NAME_TOOL(1)

NAME
       install_name_tool - change dynamic shared library install names

SYNOPSIS
       install_name_tool  [-change  old  new  ]  ... [-rpath old new ] ... [-add_rpath new ] ... [-delete_rpath new ] ... [-id
       name] file

[...]
       -rpath old new
              Changes  the  rpath  path name old to new in the specified Mach-O binary.  More than one of these options can be
              specified.  If the Mach-O binary does not contain the old rpath path name in a specified -rpath it is an  error.

```

gonuke:
Thanks @bam241

PR #114, opened 2021-08-09
added depth of 1 to git clones

shimwell:
This is a tiny PR that aims to speed up the GitHub Action builds

Currently we are git cloning the entire branch for 3 repos.

This PR changes to clone with a depth of 1 which just performs a show clone

I'm hoping this will speed up the build process fractionally

We can compare this GitHub Action times to this last run to see if the reduced download makes much impact https://github.com/svalinn/Cubit-plugin/actions/runs/1091685594

### DIFF
--- a/scripts/unix_share_build.sh
+++ b/scripts/unix_share_build.sh
@@ -110,7 +110,7 @@ function build_moab() {
     cd ${PLUGIN_ABS_PATH}
     mkdir -pv moab/bld
     cd moab
-    git clone https://bitbucket.org/fathomteam/moab -b Version5.1.0 --depth 1
+    git clone https://bitbucket.org/fathomteam/moab -b 5.3.0 --depth 1
     cd moab
     # patching MOAB CMakeLists.txt to use default find(HDF5)
     $SED -i "s/HDF5_MOAB/HDF5/" CMakeLists.txt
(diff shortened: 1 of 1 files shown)


scripts/unix_share_build.sh
```diff
@@ -110,7 +110,7 @@ function build_moab() {
     cd ${PLUGIN_ABS_PATH}
     mkdir -pv moab/bld
     cd moab
-    git clone https://bitbucket.org/fathomteam/moab -b Version5.1.0
+    git clone https://bitbucket.org/fathomteam/moab -b Version5.1.0 --depth 1
```
gonuke: Shouldn't we update this to 5.3.0 while we're at it?

gonuke (changes requested):
Can we update MOAB version, as well?

scripts/unix_share_build.sh
shimwell: Thanks, I've updated to 5.3.0 now, I just wanted to compare like for like with the timings

shimwell:
Timings appear to vary from run to run but comparing the last merge to develop which took 19m 48s with this one which takes 17m 47s I think we are seeing a small speed up in time as a result of shallow cloning

shimwell:
The last GH actions build was queuing for a while so please ignore the timings for that one. Over all this looks faster (more so when run locally than of GH actions)

gonuke:
Thanks @Shimwell

Issue #112, opened 2021-08-04
Cubit plugin in a Dockerfile glibc version error


shimwell:
Let me know if this should be opened on the MOAB repo but I've stumbled across an error when installing the plugin within a few Dockerfiles

It appears that ```GLIBC_2.29``` is required by MOAB but this is not always available in the OS
continuumio/miniconda3:4.10.3 was published 12 days ago and comes with ```GLIBC_2.28``` and continuumio/miniconda3:4.9.2 which is the previous release also has ```GLIBC_2.28``` 

FROM continuumio/miniconda3:4.9.2
![Screenshot from 2021-08-04 21-05-03](https://user-images.githubusercontent.com/8583900/128250869-9deab3b9-9470-4f28-93bc-beeaa0d6da6a.png)

FROM continuumio/miniconda3:4.10.3 with a python 3.8 environment created as Cubit appears [to not support python 3.9](https://forum.coreform.com/t/python-3-9-support-in-cubit/1199)
![Screenshot from 2021-08-04 21-30-59](https://user-images.githubusercontent.com/8583900/128251032-3959bc28-bc76-4306-9984-09a41b4cb95d.png)

ubuntu:21.04 docker image also fails as it has glibc version 2.28-10

Wondering if anyone knows a work around for this

bam241:
The problem might be related to your LD_LIBRARY_PATH, when running from command line I have to unset it....

shimwell:
Had a quick go with setting the LD_LIBRARY_PATH to nothing. Looks like the same error occurs

this is with ubuntu 21.04 base image

![Screenshot from 2021-08-04 23-39-21](https://user-images.githubusercontent.com/8583900/128264564-41b5777f-dc3e-4c7a-bf99-a9ee060c592e.png)

shimwell:
I'm thinking of adding a matrix to the build for different images as we have a few different glic versions

ubuntu 20.04 makes use of glibc 2.31
ubuntu 21.04 makes use of glibc 2.33
debian 10.10 makes use of glibc 2.28

shimwell:
Just managed to get the Cubit plugin building for Debian 10.10
https://github.com/Shimwell/Cubit-plugin/runs/3279990398?check_suite_focus=true

shimwell:
PR #115 should solve this issue

shimwell:
Thanks everyone for helping with the PR #115 that solved this problem I was having.

PR #115, opened 2021-08-09
Building ubuntu and debian inside docker

shimwell:
Following on from the last DAGMC meeting here is a PR that attempts to build the plugin for debian 10.10

The approach taken is to use a docker container to build linux flavours (ubuntu 21.04, ubuntu 20.04, debian 10.10) and to use the base github actions (without a docker container) to build for mac os.

This appears to work given the constraints:
- containerization of mac is not currently possible

This appears to help with #112 as the plugin can be build for a few more OS and OS versions.


shimwell:
Everything appears to be building 🎉 thanks @bam241 for getting this started

.github/workflows/unix_linux.yml
```diff
@@ -18,23 +18,49 @@ env:
...
+          if [ "${{ matrix.os }}" == "ubuntu" ]; then
+            apt-get update
+            apt-get install -y software-properties-common
+            add-apt-repository ppa:git-core/ppa
+            apt-get update
+            apt-get install -y git
+          elif [ "${{ matrix.os }}" == "debian" ]; then
+            apt update
+            apt-get install -y software-properties-common
+            apt update
```
gonuke: Is this update necessary?

.github/workflows/unix_linux.yml
```diff
@@ -104,6 +120,7 @@ jobs:
         run: |
           mkdir -p $FOLDER_PKG
           cd ${FOLDER_PKG}
+          apt-get install -y curl
```
gonuke: Can we add this to the earlier apt-get?

.github/workflows/unix_linux.yml
```diff
@@ -57,26 +83,16 @@ jobs:
             HDF5_PATH=/usr/local/HDF_Group/HDF5/1.12.0
           fi
 
-          if [[ "${{ matrix.os }}" == "ubuntu"*  ]]; then
+          if [ "${{ matrix.os }}" == "debian" ] || [ "${{ matrix.os }}" == "ubuntu" ]; then
             SUFFIX=Lin64
             EXT=deb
             echo "SED=sed" >> $GITHUB_ENV
-            echo "SUDO=sudo " >> $GITHUB_ENV
```
gonuke, line 64: Why don't we want SUDO defined?

.github/workflows/unix_linux.yml
```diff
@@ -57,26 +83,16 @@ jobs:
             HDF5_PATH=/usr/local/HDF_Group/HDF5/1.12.0
           fi
 
-          if [[ "${{ matrix.os }}" == "ubuntu"*  ]]; then
+          if [ "${{ matrix.os }}" == "debian" ] || [ "${{ matrix.os }}" == "ubuntu" ]; then
```
gonuke: Do we need a conditional still if we remove Mac? Or we add some failure path if we don't match this condition

gonuke (changes requested):
A few changes to clean things up.

At some point we should review the overall design to increase reuse of things rather than maintaining different scripts.

.github/workflows/unix_linux.yml
shimwell: I'm not actually sure about this part, it was in the original script.

gonuke: We probably still want it

shimwell: ok, we can leave this in for now, possibly come back to it one day

.github/workflows/unix_linux.yml
shimwell: now added to earlier apt-get

.github/workflows/unix_linux.yml
shimwell: now removed this condition. Perhaps we will bring this back one day it more OS get added

.github/workflows/unix_linux.yml
shimwell: reworked this section to reduce number of updates

shimwell:
I've addressed all the suggestions and done as recommended, thanks for the review. Let's see how it does with the CI building process.

shimwell:
> A few changes to clean things up.
> 
> At some point we should review the overall design to increase reuse of things rather than maintaining different scripts.

Sounds like a good plan, I think once GH actions allows containers of Mac OS there is some recombination that can be done

bam241:
There some issues with the plug-in artifacts uploads: https://github.com/svalinn/Cubit-plugin/actions/runs/1114012215
And:

https://github.com/svalinn/Cubit-plugin/actions/runs/1114012218

One also can test those individually to ensure they are working as intended :)

bam241:
I meant test them locally

bam241:
> > A few changes to clean things up.
> 
> > 
> 
> > At some point we should review the overall design to increase reuse of things rather than maintaining different scripts.
> 
> 
> 
> Sounds like a good plan, I think once GH actions allows containers of Mac OS there is some recombination that can be done

This is very unlikely to happen as it will broken general usage conditions of MacOS: you can only use macOS on a Mac machine... 
I.e. virtualization of macOS is only allowed on a Mac host.... allowing it on docker will imply requiring a lot of safeguards on the docker implementation side to make sure you cannot run a macOS is on a non-Mac host...
Multi-platform virtualizer apps tended to stay away from those complication in the past

.github/workflows/unix_linux.yml
```diff
@@ -0,0 +1,166 @@
...
+      - name: Prepare package
+        shell: bash -l {0}
+        run: |
+          ${system}_build_plugin_pkg ${{ matrix.cubit }}
+
+      - if: github.event_name != 'release'
+        name: Upload artifact for CI
+        uses: actions/upload-artifact@v2
+        with:
+          name: svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
```
bam241, line 152: those artifact names needs to mach the one produced by the `${system}_build_plugin_pkg` macro.
I think the macro need to be tweaked to accept `os` and `os_version` variables.

.github/workflows/unix_linux.yml
```diff
@@ -0,0 +1,166 @@
...
+          path: ${{ github.workspace }}/release/svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
```
bam241, line 153:
```suggestion
          path: ${{ github.workspace }}/release/svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
          if-no-files-found: error # 'warn' or 'ignore' are also available, defaults to `warn` 
```

bam241: this will allow to be aware when we have a problem founding the tgz file

shimwell: Thanks, this is a nice check.

.github/workflows/unix_linux.yml
shimwell: Thanks, I see the ```unix_share_build.sh``` current finds the linux (major) version using ```LINUX_VERSION=$(lsb_release -rs |cut -d"." -f1)``` . Should I remove the ```lsb_release``` part and pass in ```os_version``` as predefined environmental variable

.github/workflows/unix_mac.yml
```diff
@@ -144,14 +126,14 @@ jobs:
         name: Upload artifact for CI
         uses: actions/upload-artifact@v2
         with:
-          name: svalinn-plugin_${{ matrix.os }}_cubit_${{ matrix.cubit }}.tgz
-          path: ${{ github.workspace }}/release/svalinn-plugin_${{ matrix.os }}_cubit_${{ matrix.cubit }}.tgz
+          name: svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
+          path: ${{ github.workspace }}/release/svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
```
bam241, line 130:
```suggestion
          path: ${{ github.workspace }}/release/svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
          if-no-files-found: error # 'warn' or 'ignore' are also available, defaults to `warn` 
```

bam241 (changes requested):
this change would be required for Mac as well :)

scripts/unix_share_build.sh
```diff
@@ -25,19 +24,26 @@ function mac_install_prerequisites() {
...
 }
 
 function linux_install_prerequisites() {
     TZ=America/Chicago
     $SUDO ln -snf /usr/share/zoneinfo/$TZ /etc/localtime
     $SUDO sh -c 'echo $TZ > /etc/timezone'
     $SUDO apt-get update -y
-    $SUDO apt-get install -y g++ libeigen3-dev patchelf git cmake curl lsb-release python3  lsb-core
+    if [ "$OS" == "ubuntu" ]; then
+    $SUDO apt-get install -y g++ libeigen3-dev patchelf git cmake curl lsb-release python3 lsb-core
```
gonuke, line 39: Notice that this script relies on `SUDO` but it never gets set.  Current behavior is probably that this resolves to nothing and doesn't matter... but we should either remove the `$SUDO` here, or set the variable as appropriate.

gonuke (changes requested):
Something happened to SUDO

.github/workflows/unix_linux.yml
shimwell: Went for the option of passing in another variable called os_version in #116

scripts/unix_share_build.sh
shimwell: I've now set SUDO to an empty string in the mac and linux github actions scripts in commit 6dc6a66

shimwell: Ah that setting of SUDO to an empty string appears to have not worked

shimwell:
Looks like Trelis 17.1.0 is failing on Ubuntu 21.04 due to not finding HDF5

```Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS) (found  version "")```

This was not actually build previously so I shall add it to the exclude list.

Perhaps it is fair to exclude combinations of old cubit with new os or vice versa

Ubuntu 20.04 with Cubit 17.0.1 was previously excluded, so I think it is reasonable to also exclude 21.04 with this version of Cubit

**Comment by shimwell** [comment]
Builds are passing for all the combinations 🎉

File: .github/workflows/unix_linux.yml [hunk 9]
```diff
@@ -0,0 +1,163 @@
...
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+jobs:
+  building-plugin:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        cubit: [2021.4, 2021.5]
+        os: [ubuntu]
+        os_version: [20.04, 21.04]
```
gonuke, line 27: Can we add in 18.04 here?

gonuke (changes requested):
Just wondering about another Ubuntu version for building?

.github/workflows/unix_linux.yml
shimwell: Certainly

shimwell: Ah looks like 18.04 is failing with ```Could NOT find HDF5 (missing: HDF5_LIBRARIES HDF5_INCLUDE_DIRS) (found  version "")```

gonuke: Looking at the currently working version... it appears that we didn't support 18.04 with anything other than Cubit 17.  @bam241 will remember why...  You can remove it again.  I think this new version (without 18.04 in the matrix) will build at least as many combinations as the old version.

shimwell: Yep I think this new version builds everything the old version did and also a 4 extras
Ubuntu 21.04 Cubit 2021.4
Ubuntu 21.04 Cubit 2021.5
Debian 10.10 Cubit 2021.5
Mac OS 10.15 Cubit 2021.5

shimwell: |              | Trelis 17.01       | Cubit 2021.4 | Cubit 2021.5 |
|--------------|--------------------|--------------|--------------|
| Ubuntu 18.04 | :heavy_check_mark: |              |              |
| Ubuntu 20.04 |                    |  :heavy_check_mark:|  :heavy_check_mark:|
| Ubuntu 21.04 |                    |  :heavy_check_mark:|  :heavy_check_mark:|
| Debian 10.10 |                    |              |  :heavy_check_mark:|
| Mac OS 10.15 |  :heavy_check_mark:|  :heavy_check_mark:|  :heavy_check_mark:|

gonuke:
Thanks @Shimwell - this gives us a good starting point to refine and grow

PR #116, opened 2021-08-10
added os_version to unix build macro

shimwell:
This PR into a feature branch is one option for getting ```os_version``` from the github action into the unix_share_build script as suggested in the review for #115

The only down side I can see with this approach is that the unix_share_build script currently makes use of ```lsb_release``` to find the Linux version. So perhaps this should also be removed?

### DIFF
--- a/scripts/unix_share_build.sh
+++ b/scripts/unix_share_build.sh
@@ -12,6 +12,7 @@
 # - SUDO : specify the command prefix to run as root (if any)
 # - SED : command to run sed (usually sed for linux, gsed for mac)
 # - OS : reference to the Operating system, used to name the plugin tarball (and when using setup_var() function)
+# - OS_VERSION : reference to the Operating system version, used to name the plugin tarball (and when using setup_var() function)
 
 set -ex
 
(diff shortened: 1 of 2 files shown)


scripts/unix_share_build.sh
```diff
@@ -271,9 +272,9 @@ function linux_build_plugin_pkg(){
     cd ..
     ln -sv svalinn/libsvalinn_plugin.so .
     cd ../..
-    tar --sort=name -czvf svalinn-plugin_${OS}_cubit_$1.tgz bin
-    chmod 666 svalinn-plugin_${OS}_cubit_$1.tgz
-    cp svalinn-plugin_${OS}_cubit_$1.tgz $SCRIPTPATH/release/
+    tar --sort=name -czvf svalinn-plugin_${OS}-${OS_VERSION}_cubit_$1.tgz bin
```
gonuke: It probably makes sense to assign this tarball file name to a variable and reuse it over these three lines to prevent typos causing errors

gonuke (changes requested):
This seems like a good plan to me - just one suggestion for an improvement of an inherited issue...

gonuke:
I think it makes sense to rely on the `OS_VERSION` variable instead of the `lsb_release` command, but I wonder how we avoid an inconsistency?

shimwell:
> I think it makes sense to rely on the `OS_VERSION` variable instead of the `lsb_release` command, but I wonder how we avoid an inconsistency?

We could use ```lsb_release``` to check the user specified / github action specified argument. If there is disagreement then the script could raise an error

I was also half thinking of having ```os_minor_version``` and ```os_major_version``` in the github action but that can wait for iteration 2 :-)

shimwell:
Looks like we have settled on an option for this, so I shall merge it into the other feature branch so that the CI can be triggered

PR #117, opened 2021-08-12
add mac os for all cubit versions

shimwell:
While chatting with @andergray we noticed Cubit Plugin was not available for Mac OS 11

It looks like Mac OS 11 is available as a runner so I thought it is worth trying to build on that OS

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

### DIFF
--- a/.github/workflows/unix_mac.yml
+++ b/.github/workflows/unix_mac.yml
@@ -24,7 +24,7 @@ jobs:
       matrix:
         cubit: [17.1.0, 2021.4, 2021.5]
         os: [macos]
-        os_version: [10.15]
+        os_version: [10.15, 11]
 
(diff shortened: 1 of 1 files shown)


shimwell:
Ah sorry I just noticed ``` The macOS 11 virtual environment is currently provided as a private preview only``` so this PR can be parked till that OS becomes more widely available.

bam241:
I am not sure this is necessary. So far we have build on only 1 macOS version but the build were tested and used on multiples (at least 2 that I now of...)

Maybe the only things to do is to rename the tarball

shimwell:
The existing Mac OS plugin has been tested on 10.15 and 11 to some extent and therefore has been renamed in PR #121 which mainly negates the need for this PR

PR #122, opened 2021-08-26
Add exclusion path to limit the extend of our CI, to only targeted file changes

bam241:
this will exclude change not on the code or the corresponding GitHub Action workflow to trigger some part of the CI.

Docs change are now excluded from workflows,
Linux workflow changes will only trigger Linux CI same goes for MacOS...

### DIFF
--- a/.github/workflows/unix_linux.yml
+++ b/.github/workflows/unix_linux.yml
@@ -6,11 +6,21 @@ on:
   pull_request:
     branches:
       - develop
-      - master
+    paths-ignore:
+      - '.github/workflows/windows.yml'
+      - '.github/workflows/unix_linux.yml'
+      - '*.md'
+      - 'LICENSE'
+
   push:
     branches:
       - develop
-      - master
+    paths-ignore:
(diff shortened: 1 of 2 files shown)


.github/workflows/unix_linux.yml
```diff
@@ -6,11 +6,21 @@ on:
   pull_request:
     branches:
       - develop
-      - master
+    paths-ignore:
+      - '.github/workflows/windows.yml'
+      - '.github/workflows/unix_linux.yml'
```
gonuke: Shouldn't this one be ignoring Mac and the Mac one ignoring Linux? Seems backwards

gonuke: Shouldn't this one be ignoring Mac and the Mac one ignoring Linux? Seems backwards

gonuke (changes requested):
Looks like some workflow ignores are backwards

gonuke:
Thanks @bam241 - merging now to give some CI relief

PR #123, opened 2021-09-22
trying to download and install the plugin

bam241:


test_plugin/test.jou
```diff
@@ -1,5 +1,5 @@
 help mcnp
 help dagmc
 import mcnp test.i
```
pshriwise: Per the DAGMC meeting this morning:

```suggestion
import mcnp "/__w/Cubit-plugin/Cubit-plugin/test_plugin/test.i"
```

pshriwise:
Tests are passing!! 🎉 

Whatcha think @bam241? Is this ready for review?

bam241:
depends how deep we want to implement the test.

I was only playing with ubuntu 20.04 and cubit 2021.5

Not sure about the when and how, but we would need to extend this to:
- [ ]  Mac 
- [ ] other version of linux/ubuntu
- [ ] windows 
- [ ] additional version of cubit ?

bam241:
Anyhow, there is some cleanup needed, I commented out all the matrix on the build side for speed purposes, but those need to be back at least on the build side.

of the test side, if we enable the full matrix, some tweaks have to be done with some path names (I think I was directly calling the cubit 2021.5 folder....)

.github/workflows/unix_linux.yml
```diff
@@ -32,20 +32,22 @@ jobs:
...
+        #   - os: ubuntu
+        #     os_version: 18.04
+        #     cubit: 17.1.0
 
-          - os: debian
-            os_version: '10.10'  # using a 'string' here as the 0 gets rounded away otherwise
-            cubit: 2021.5
+        #   - os: debian
+        #     os_version: '10.10'  # using a 'string' here as the 0 gets rounded away otherwise
+        #     cubit: 2021.5
```
bam241, line 48: One need to revert my changes here...s

.github/workflows/unix_linux.yml
```diff
@@ -171,3 +172,130 @@ jobs:
...
+          tar -xzvf /__w/Cubit-plugin/Cubit-plugin/release/svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
+
+
+      - name: Test the plugin (Python)
+        run: |
+          cd
+          git clone https://github.com/Shimwell/cubit_plugin_testing.git
+          cd cubit_plugin_testing
+
+          python stp_to_h5m_cubit_script.py
```
bam241, line 291: I am pretty sure this will not be compatible with version other than 2021.5 (hardcoded path...)

.github/workflows/unix_linux.yml
```diff
@@ -171,3 +172,130 @@ jobs:
...
+        # os_version: [20.04, 21.04]
+        os_version: [20.04]
+        # include:
+        #   - os: ubuntu
+        #     os_version: 18.04
+        #     cubit: 17.1.0
+
+        #   - os: debian
+        #     os_version: '10.10'  # using a 'string' here as the 0 gets rounded away otherwise
+        #     cubit: 2021.5
```
bam241, line 194: to re-enable for the full matrix testing.

bam241 (commented):
here are some quick thoughts about what remain to be done in here....

PR #125, opened 2021-09-30
Changing MCNP import syntax to lowercase.

pshriwise:
Windows users have reported that the MCNP import command only works when using an uppercase "MCNP". This changes the syntax to allow a lowercase name in the input command for consistency between Windows and Linux as well as consistency in the input syntax between the importer and the exporter.


gonuke:
Thanks @pshriwise

Issue #129, opened 2021-12-15
Build Cubit for Coreform Cubit version 2021.11


shimwell:
Coreform[ release Cubit 2021.11](https://coreform.com/products/downloads/) recently and we don't yet to have a plugin built for this version.

I believe that it is fairly straight forward to update the Github Action scripts so that the Svalinn Plugin for this new version of Cubit can be included in the next release.

shimwell:
The release action for linux appears to be failing when servers are not responding

I've rerun it 3 times in cases it is an unlucky glitch but still different individual runs are failing with this error

```
Run svenstaro/upload-release-action@v2
/usr/bin/docker exec  54c166f3dcbdceb6890607dd672a6352ec897d6ad00bc8a7a5f9cc6bbca66813 sh -c "cat /etc/*release | grep ^ID"
Error: request to https://uploads.github.com/repos/svalinn/Cubit-plugin/releases/55456771/assets?name=svalinn-plugin_ubuntu-21.04_cubit_2021.4.tgz& failed, reason: read ECONNRESET
```

makeclean:
Does it matter that this is pointing to version 2021.4 rather than 2021.11?

shimwell:
Oh that is just the specific version that failed for that github action run, it builds 9 different combinations on the linux yml.

It would be ideal if they all build and get uploaded. I know we have the older 2021.4 on the previous release but it would be nice to get them all for each release

PR #130, opened 2021-12-15
Adding cubit 2021.11

shimwell:
Coreform[ release Cubit 2021.11](https://coreform.com/products/downloads/) recently as mentioned in #129 

This PR adds this version to the matrix of build options and other places in the build scripts so that the plugin can be built for this new version

It appears to build without and version specific additions to the scripts and I have built it over on this fork branch 

I have made a test release on separate fork
https://github.com/shimwell/Cubit-plugin/releases

Release builds performed on Github Actions are the top 3 in this list
https://github.com/shimwell/Cubit-plugin/actions

The linux one appears to fail sometimes, so that is the reason I've added ```overwrite:true``` to the artifact upload. They the action ca be rerun for the same release.


gonuke:
LGTM - thanks @shimwell

PR #131, opened 2021-12-17
moving to maintained action and splitting up linux github actions yml and 

shimwell:
The existing Github Actions scripts make use of upload-release-asset which is now no longer maintained and suggests using action-gh-release instead.

So I think it might be time to change the GitHub Action used to upload the compressed files to the release.

Additionally as mentioned in the call today the combined action for Ubuntu and Debian upload was getting a bit large so I've split those two into separate yml files which.

The github actions triggered by a test release on my fork all worked as the failing option Ubuntu 21.04 and Cubit 2021.4 has been removed from the matrix.
https://github.com/shimwell/Cubit-plugin/actions

### DIFF
--- a/.github/workflows/linux_ubuntu.yml
+++ b/.github/workflows/linux_ubuntu.yml
@@ -34,13 +34,17 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        cubit: [2021.4, 2021.5, 2021.11]
+        cubit: [2021.5, 2021.11]
         os: [ubuntu]
         os_version: [20.04, 21.04]
         include:
           - os: ubuntu
             os_version: 18.04
             cubit: 17.1.0
+        include:
+          - os: ubuntu
+            os_version: 20.04
+            cubit: 2021.4
 
     name: 'Cubit ${{ matrix.cubit }} Build for ${{ matrix.os }} ${{ matrix.os_version }} of Svalinn Plugin'
 
--- a/.github/workflows/linux_ubuntu.yml
+++ b/.github/workflows/linux_ubuntu.yml
@@ -1,4 +1,4 @@
-name: Linux builds
+name: Linux Ubuntu builds
 
 on:
   # allows us to run workflows manually
@@ -9,6 +9,7 @@ on:
     paths-ignore:
       - '.github/workflows/windows.yml'
       - '.github/workflows/unix_mac.yml'
+      - '.github/workflows/linux_debian.yml'
       - '*.md'
       - 'LICENSE'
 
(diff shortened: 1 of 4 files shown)


gonuke:
Is the combination of Ubuntu 21.04 and Cubit 2021.4 known to fail before now?

.github/workflows/linux_ubuntu.yml
```diff
@@ -171,12 +166,8 @@ jobs:
           path: ${{ github.workspace }}/release/svalinn-plugin_${{ matrix.os }}-${{ matrix.os_version }}_cubit_${{ matrix.cubit }}.tgz
           if-no-files-found: error # 'warn' or 'ignore' are also available, defaults to `warn` 
     
-      - if: github.event_name == 'release'
-        name: Upload binaries into the release
-        uses: svenstaro/upload-release-action@v2
+      - name: Upload binaries into the release
+        uses: softprops/action-gh-release@v1
+        if: startsWith(github.ref, 'refs/tags/')
```
gonuke, line 171: Is there a substantive reason to change the condition?  Why not put the conditional first?

.github/workflows/linux_ubuntu.yml
```diff
@@ -32,23 +34,18 @@ jobs:
...
             cubit: 17.1.0
 
-          - os: debian
-            os_version: '10.10'  # using a 'string' here as the 0 gets rounded away otherwise
-            cubit: 2021.5
-
-          - os: debian
-            os_version: '10.10'  # using a 'string' here as the 0 gets rounded away otherwise
-            cubit: 2021.11
+          - os: ubuntu
```
gonuke, line 45: this has now excluded the U21.04 + C2021.4 combination - was that failing before?

gonuke (changes requested):
Thanks for the work here @shimwell - I wonder how to deal with third party actions becoming unsupported in the long term?

.github/workflows/linux_ubuntu.yml
shimwell: That is just how the example suggested to do it. I think that the action handles the condition instead of the yml
https://github.com/softprops/action-gh-release

.github/workflows/linux_ubuntu.yml
shimwell: I think it was not failing before. I can't quite figure out why it is failing now.

> Thanks for the work here @shimwell - I wonder how to deal with third party actions becoming unsupported in the long term?

I think the newer replacements are usually easier to use which I guess is part of the reason they become the preferred option. They also tend to offer the same or better functionality. So perhaps upgrading actions won't be much trouble. This action repo was kind enough to recommend  a replacement one.

shimwell: > this has now excluded the U21.04 + C2021.4 combination - was that failing before?

I don't quite understand why it no longer works but it seems that U21.04 + C2021.4 was working previously. I can manually upload it to the release of that helps

PR #135, opened 2022-07-21
adding 2022.6 version on build workflow

bam241:
adding 2022.6 in the version stack


bam241:
conclusion of the quick test:
Linux build seems to work for all version
MacOS workflow is out-of-date. we need to update it be faire testing anything
Windows seems to have some issue with building MOAB, this might only be related to 2022.6, I'll run more test to figure this out

bam241:
> Windows seems to have some issue with building MOAB, this might only be related to 2022.6, I'll run more test to figure this out

All windows workflows are broken on MOAB build

.github/workflows/unix_mac.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     runs-on: macos-10.15
```
gonuke: Unless there are other failures, this should be enough to fix the mac builds
```suggestion
    runs-on: macos-11
```

gonuke:
Mac is failing because of a deprecated call to `finite()`.  MOAB tries to invoke the correct instance of this function, given that apparently many exist - see MOAB's `Util.hpp`.

It appears that we may not have access to the `MOAB_HAVE_ISFINITE` macro, either because we are not accessing the file that sets this (less likely because other things would fail) or that our CMake build of MOAB fails to set this in the config file we do access.  This may need MOAB updates.

bam241:
I don't believe this is the right diagnostic.
I think you were looking at the build that was cancelled because the build 2021.11 failed:
```
Run ${system}_build_plugin_pkg 2021.11
[21](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:22)
+ mac_build_plugin_pkg 2021.11
[22](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:23)
+ cd /Users/runner/work/Cubit-plugin/Cubit-plugin/..
[23](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:24)
+ mkdir -p pack/MacOS/plugins/svalinn
[24](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:25)
+ cd pack/MacOS/plugins/svalinn
[25](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:26)
+ cp -pPv /Users/runner/work/Cubit-plugin/Cubit-plugin/../lib/libiGeom.dylib /Users/runner/work/Cubit-plugin/Cubit-plugin/../lib/libmcnp2cad.dylib /Users/runner/work/Cubit-plugin/Cubit-plugin/../lib/libsvalinn_plugin.so .
[26](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:27)
/Users/runner/work/Cubit-plugin/Cubit-plugin/../lib/libiGeom.dylib -> ./libiGeom.dylib
[27](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:28)
+ cp /usr/local/opt/szip/lib/libsz.2.dylib .
[28](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:29)
/Users/runner/work/Cubit-plugin/Cubit-plugin/../lib/libmcnp2cad.dylib -> ./libmcnp2cad.dylib
[29](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:30)
/Users/runner/work/Cubit-plugin/Cubit-plugin/../lib/libsvalinn_plugin.so -> ./libsvalinn_plugin.so
[30](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:31)
cp: /usr/local/opt/szip/lib/libsz.2.dylib: No such file or directory
[31](https://github.com/svalinn/Cubit-plugin/runs/7456715236?check_suite_focus=true#step:11:32)
Error: Process completed with exit code 1.
```

the other ones stopped with : 
`Error: The operation was canceled.`

bam241:
I think one needs to investigate this:
`cp: /usr/local/opt/szip/lib/libsz.2.dylib: No such file or directory`

bam241:
Mac build works, one need to test it.
Linux build is failing, because of some connection issue, I don't think that comes from us.

scripts/unix_share_build.sh
```diff
@@ -230,6 +230,7 @@ function linux_setup_cubit() {
 function build_plugin(){
     cd ${PLUGIN_ABS_PATH}
     cd Cubit-plugin
+    git config --global --add safe.directory /__w/Cubit-plugin/Cubit-plugin
```
gonuke:
```suggestion
    git config --global --add safe.directory ${GITHUB_WORKSPACE}
```

gonuke (changes requested):
Thanks @bam241! One small change that I think makes thinks more robust, and then we can prob merge this.

bam241:
@gonuke what do we want to do with the Windows build ?

it seems that HDF5 is not playing nicely with MOAB any more....
It would probably be a quick/easy fix, but I don't have the hardware to test it anymore....

bam241:
@gonuke what is the plan there ?

Do we want to include the windows fix in there as well ?

gonuke:
I think we can go ahead without the windows fix since we don't have a good place to develop that at the moment

gonuke:
Since I just merged @shimwell 's PR with the Readme table, can you rebase and update that, please?

gonuke (approved):

gonuke:
Actually, maybe this doesn't need a rebase

gonuke:
Merging this now - the updated table in the Readme should come with a release.

PR #138, opened 2022-08-18
Add badges to table

gonuke:
Here is what it would look like with badges, including older versions in a different color.


shimwell (approved):
This looks much nice in my opinion.

Happy to merge this in if there are no objections

PR #139, opened 2022-09-29
Add information about Sandia Cubit compatibility

gonuke:
Based on info from Coreform, adding info about which Sandia Cubit version works.

Also anticipates next release...???

See [here](https://github.com/gonuke/Cubit-plugin/blob/sandia/README.md) for updated README


pshriwise:
Thanks for this update, @gonuke! Nice to have some more info on the connection between these Cubit flavors.

@jbae11 has volunteered to try the plugin packages out for Sandia version 15.8 if you haven’t already.

jbae11:
I've requested 15.8 and will probably get it next week, the poc is in travel.

jbae11:
@pshriwise I might be understanding it wrong, but it looks like all the ubuntu version (except 18.04) have been tested with 15.8? So is it just the mac that I need to test?

pshriwise:
> @pshriwise I might be understanding it wrong, but it looks like all the ubuntu version (except 18.04) have been tested with 15.8? So is it just the mac that I need to test?

I think that's right, yep!

pshriwise:
@jbae11 I think that's right but I don't see a link to a Mac version for v0.2.4. In fact, I don't see any of those artifacts on the [release page for v0.2.4](https://github.com/svalinn/Cubit-plugin/releases/v0.2.4). I can download them using the links in the README though.

@gonuke any thoughts here?

jbae11:
@pshriwise I can only find the versions up to [v0.2.3](https://github.com/svalinn/Cubit-plugin/releases) in the release page, is there another link in the README that has v0.2.4?

pshriwise:
> @pshriwise I can only find the versions up to [v0.2.3](https://github.com/svalinn/Cubit-plugin/releases) in the release page, is there another link in the README that has v0.2.4?

Not for Mac that I could see. I'm not sure why though. IIRC we should just need to enable it in the automate builds.

jbae11:
okay if there isn't one pre-built, I can try to build it myself (although not sure how long that's going to take)? Let me know how you want me to proceed.

gonuke:
This is a draft PR waiting for the 0.2.4 release to happen!

gonuke:
There were issues when trying to build for this version of Cubit on MacOS.  @bam241 can perhaps remember what they are.  Also some discussion in #135

gonuke:
Actually - @bam241 fixed the MacOS problem 🎉 

You can find a build tarball here: https://github.com/svalinn/Cubit-plugin/actions/runs/2883295348

pshriwise:
> This is a draft PR waiting for the 0.2.4 release to happen!

Welp, that would explain it. 😆Sorry, I'm a little out of date with respect to the development cycle here. 

@jbae11 if you can give the link @gonuke provided a try that would be great!

jbae11:
Thanks @gonuke @pshriwise @bam241 

I untared, moved the files, but it seems like Cubit is not recognizing the plugin (i.e. `help dagmc` errors). Is there an environment variable I should be setting or something?


image

gonuke:
Thanks @jbae11 - I am still not entirely clear where files need to be placed for the Sandia version of Cubit.  I'll have to get that info next.

gonuke:
Also, perhaps someone with Coreform MacOS experience can advise? @bam241 ?

pshriwise:
Lookin at @jbae11's screenshot again... we ship a `.so` file with the tarball for OSX?

bam241:
> Lookin at @jbae11's screenshot again... we ship a `.so` file with the tarball for OSX?

we do ship both, `.dylib` for either static or shared (can't recall which is which), and `.so` for the other.

I think the problem here is likely a path problem.

We build the plugin to be install in `Coreform-Cubit-X.Y.app/Contents/MacOS/` the path is not ready for the `Sandia` version as it does not match exactly the `Coreform` one.

Sadly one did not figure out out to deal with relative path rather than absolute path....


sorry it took me so long to see the notification....

bam241:
one solution would be to build a sandia specific version with the proper sandia path.

this might only require to build a different copy using the correct Sandia path when setting them before packing in [mac_build_plugin_pkg(){](https://github.com/svalinn/Cubit-plugin/blob/1da8c0ccfd045b9f4bd10f2aa3d5bcd541b9590a/scripts/unix_share_build.sh#L282)

bam241:
looking at it more carefully, we actually only correct the path for 17.1.


But I believe the same kind of "touch" would be required to build a proper Sandia version from the one we build

jbae11:
so should I wait for another build or is there a way for me to alter that path on my end?

jbae11:
Also I'm not sure if this is a useful datapoint, but I tried the same thing with Cubit15.8 on Ubuntu 20.04, and it didn't work initially, but worked when I added the `..../Cubit-15.8/bin` and `..../Cubit-15.8/bin/plugins/svalinn` to the `LD_LIBRARY_PATH`. I tried the same thing with the mac, but still no luck.

gonuke:
While we're still working out the installation details.... does it makes sense to merge this since the release page includes this info?

PR #140, opened 2022-10-07
add Cubit 22.4 for Debian

gonuke:
Add Cubit 2022.4 for Debian to complete the linux builds for a new release.


bam241 (approved):
This looks good to me.

Maybe eventually we should create a input file that contains all the combinaison of OS/Cubit version that would be parse and run in CI.

I am not sure this is possible but it would be simpler and more centralized....

shimwell (approved):
Super to see how easy it has become to add a new version. Great to see the ci working first time

Note we will need a release to distribute the new package

PR #142, opened 2022-12-01
Add cubit 2023.6-dev to build

gonuke:
First attempt to add 2023.1-rc to linux.  Skipping 2022.11 because it has no SDK files.


gonuke:
It appears that the SDK files are missing from the tarball.  I've [raised this on the forum](https://forum.coreform.com/t/sdk-files-missing-in-2022-11-tarball/1740).

gonuke:
Trying to resolve SDK errors on the linux side: https://forum.coreform.com/t/wrong-location-of-libcf-license-server-so-in-sdk/1885/3

Issue #143, opened 2022-12-09
Does the plugin work for OSX 12, 13?


veeshy:
I see the plugin builds for osx 11, is it _supposed_ to work for other versions? It didn't work for me, but there weren't any lib errors, cubit simply didn't have the `dagmc` cmd available after copying the files into the `MacOS/plugins/` folder.

Perhaps related but the plugin dir in osx (left) and linux (right) builds has different files, is that expected?
<img width="625" alt="image" src="https://user-images.githubusercontent.com/5826161/206812564-c5d0c3e5-e02f-4943-aca6-78cf70c576dc.png">

bam241:
it should be ok, 

I am unsure it works with a M1 processor. (It is compile for Mac-intel, but might work with rosetta 2..)

you need to exact same version of the plugin than the cubit you are using.

can you tell me which version you are attempting to use so I can give it a try ?

bam241:
I just confirmed that the 2021.11 plugin work with MacOS 13.0.1 on a MAC-intel.

@veeshy please let me know if you need me to check other build (or a M1-MAC).

veeshy:
I’m on intel, 22.4 and downloaded the correct versions. Will try with 21.11 tomorrow just in case (edit: did not work either). Thanks for checking. 

Also might be my fault in multiple ways, do you need dagmc compiled first or is the cubit plugin standalone?

veeshy:
Looked a little more at this, I see when using the cubit no gui cli the plugin load errors with:

```
Loading command plugins from /Applications/Coreform-Cubit-2022.4.app/Contents/MacOS/plugins
[2022-12-13 22:02:04.960] [app_logger] [debug] CubitDynamicLoader: load_library_sys: /Applications/Coreform-Cubit-2022.4.app/Contents/MacOS/plugins/libmaterial_commands.so
[2022-12-13 22:02:05.084] [app_logger] [debug] CubitDynamicLoader: load_library_sys: /Applications/Coreform-Cubit-2022.4.app/Contents/MacOS/plugins/libsvalinn_plugin.so
[...]

PR #144, opened 2022-12-16
fixing macos 2022.4 plugin

bam241:
this should allow add the missing lib and to correct its @rpath in the plugin library

### DIFF
--- a/scripts/unix_share_build.sh
+++ b/scripts/unix_share_build.sh
@@ -74,6 +74,8 @@ function setup_var() {
         CUBIT_PATH="/opt/Coreform-Cubit-2021.3"
     elif [ "$1" == "2021.4" ] ; then
         CUBIT_PATH="/opt/Coreform-Cubit-2021.4"
+    elif [ "$1" == "2022.4" ] ; then
+        CUBIT_PATH="/opt/Coreform-Cubit-2022.4"
     else
         echo "unknown Cubit version"
         return 1
@@ -288,6 +290,11 @@ function mac_build_plugin_pkg(){
     cp -pPv ${PLUGIN_ABS_PATH}/lib/* .
     cp /usr/local/opt/libaec/lib/libsz.dylib .
     install_name_tool -change /usr/local/opt/libaec/lib/libsz.dylib @rpath/libsz.dylib libsvalinn_plugin.so
+
+    if [ "$1" == "2022.4" ] ; then
+    cp /usr/local/opt/libaec/lib/libsz.2.dylib .
+    install_name_tool -change /usr/local/opt/libaec/lib/libsz.2.dylib @rpath/libsz.2.dylib libsvalinn_plugin.so
+    fi
 
     # restoring correct RPATH for 17.1 (bin does not exist as it is not shipped with SDK)
     if [ "$1" == "17.1.0" ] ; then
--- a/scripts/unix_share_build.sh
+++ b/scripts/unix_share_build.sh
@@ -176,7 +176,6 @@ function remove_app_logger() {
 }
 
 function mac_setup_cubit() {
-
     cd ${FOLDER_PKG}
     hdiutil convert ${CUBIT_PKG} -format UDTO -o cubit_eula.dmg.cdr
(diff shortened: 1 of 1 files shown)


bam241:
this should fix #143 for 2022.4
cc @veeshy

once it is build you should be able to find a new version of the plugin in the build artifacts. 
For your specific version here:
https://github.com/svalinn/Cubit-plugin/actions/runs/3712499162

scripts/unix_share_build.sh
```diff
@@ -74,6 +74,8 @@ function setup_var() {
         CUBIT_PATH="/opt/Coreform-Cubit-2021.3"
     elif [ "$1" == "2021.4" ] ; then
         CUBIT_PATH="/opt/Coreform-Cubit-2021.4"
+    elif [ "$1" == "2022.4" ] ; then
+        CUBIT_PATH="/opt/Coreform-Cubit-2022.4"
```
bam241, line 78: this is not used, but I thought it might be better for consistency

bam241:
@veeshy the new build seems to be working for me 

could you confirm ?
```
(Non-Commercial License) Cubit>help export dagmc

Help for words: export & dagmc.

export dagmc <filename> [faceting_tolerance <faceting tolerance>] [length_tolerance <length tolerance>]
     [normal_tolerance <normal tolerance>] [make_watertight] [verbose]
     [fatal_on_curves]


Help for words: export & dagmc.

export dagmc <filename> [faceting_tolerance <faceting tolerance>] [length_tolerance <length tolerance>]
     [normal_tolerance <normal tolerance>] [make_watertight] [verbose]
     [fatal_on_curves]
```

veeshy:
Can confirm it works for me

bam241:
Great.

Could you take a look  @gonuke ?

scripts/unix_share_build.sh
```diff
@@ -289,6 +291,11 @@ function mac_build_plugin_pkg(){
     cp /usr/local/opt/libaec/lib/libsz.dylib .
     install_name_tool -change /usr/local/opt/libaec/lib/libsz.dylib @rpath/libsz.dylib libsvalinn_plugin.so
 
+    if [ "$1" == "2022.4" ] ; then
+    cp /usr/local/opt/libaec/lib/libsz.2.dylib .
+    install_name_tool -change /usr/local/opt/libaec/lib/libsz.2.dylib @rpath/libsz.2.dylib libsvalinn_plugin.so
+    fi
```
gonuke: What if we made the lib name or version a variable based on the Cubit version, and then use that in various lines that are common (does that make sense?)

scripts/unix_share_build.sh
gonuke: Why isn't it used? At least it avoids the exit condition, right?

gonuke (changes requested):
Thanks @bam241 - One small suggestion

scripts/unix_share_build.sh
bam241: it is only used if we run this using the build scripts and not the CI.

scripts/unix_share_build.sh
bam241: Do you mean a `$cubit_version` variable instead for using an argument ?

gonuke: Here's what I was thinking:
```suggestion
    libsz=libsz.dylib
    if [ "$1" == "2022.4" ] ; then
    libsz=libsz.2.dylib
    fi    

    cp /usr/local/opt/libaec/lib/$libsz .
    install_name_tool -change /usr/local/opt/libaec/lib/$libsz @rpath/$libsz libsvalinn_plugin.so
```

bam241:
@gonuke, sorry it took me so long.

next time, if there is just a small suggestion like that, feel free to commit it, and merge it, so this doesn't have to wait for me and my limited availability....

bam241:
ping @gonuke  ?

gonuke (approved):

gonuke:
Thanks @bam241 - LGTM!

Issue #148, opened 2023-08-29
Update README to describe plugin activation for Coreform Cubit


pshriwise:
Now that the plugin is shipping with Coreform Cubit (🎉), we should update the README to describe how to activate the plugin for Linux, Mac, and Windows.

PR #149, opened 2023-08-29
Adding plugin instructions for coreform cubit

pshriwise:
Addressing #148


gonuke (approved):
Thanks for the quick update @pshriwise

PR #157, opened 2024-06-08
Support degenerate tori with negative major radius

### LINKED ISSUE #156: We do not support the 3rd degenerate case for Tori
The mcnp manual states there are three 'kinds' of torus;
1.  major radius > minor radius and major radius > 0 (normal torus)
2.  major radius < minor radius and major radius > 0 (degenerate torus where there is some overlap forming a D shape)
3.  major radius < minor radius and major radius < 0 (degenerate torus where there is some overlap forming a lens shape)

It seems we do not support the third use case.

gonuke:
This method already supported a degenerate torus with major radius > 0, with a cross-section that includes more than 1/2 of a circle, truncated by a chord across that circle.

This should now support a degenerate torus with major radius < 0, with a cross-section that includes less than 1/2 of a circle, also truncated by a chord across that circle.

Fixes #156

### DIFF
--- a/iGeom/iGeom.cpp
+++ b/iGeom/iGeom.cpp
@@ -273,16 +273,23 @@ void iGeom_createTorus(iGeom_Instance instance,
         STRAIGHT_CURVE_TYPE, v1, v5);
     profile_edges.insert(curve);
 
-    //the first arc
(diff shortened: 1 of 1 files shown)


gonuke:
I'm not sure we have a good mechanism to test this any more, since we don't build this routinely.  Perhaps I can point someone from Coreform to look at it.

PR #160, opened 2025-06-20
Redirect users to Coreform

gonuke:
Since this project is no longer actively supported, direct users to Coreform and Sandia points of contact


pshriwise (approved):
Thanks @gonuke!

